PR #1, opened 2022-03-07
feat: Add provider

kylecarbs:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @hashicorp/terraform-provider-devex
```
coadler, line 1: :sunglasses:

kylecarbs: awko taco

PR #5, opened 2022-03-25
feat: Add auth property to coder_agent_script

kylecarbs:
This enables explicit definition of auth type, so the agent
doesn't misinterpret the running environment.

This also adds "username" and "name" properties to "coder_workspace"
to allow for pretty resource naming inside a cloud.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -65,6 +65,8 @@ func New() *schema.Provider {
 						transition = "start"
 					}
 					rd.Set("transition", transition)
+					rd.Set("username", os.Getenv("CODER_WORKSPACE_USERNAME"))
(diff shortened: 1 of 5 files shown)


internal/provider/provider.go
```diff
@@ -65,6 +65,8 @@ func New() *schema.Provider {
 						transition = "start"
 					}
 					rd.Set("transition", transition)
+					rd.Set("username", os.Getenv("CODER_WORKSPACE_USERNAME"))
```
bpmct, line 68: thoughts on `owner` or `CODER_WORKSPACE_OWNER` instead? to me, username sounds like a property unique to the workspace.

bpmct (approved):
left one comment on naming. NB

PR #8, opened 2022-04-08
feat: Remove agent_script data source to simplify resources

kylecarbs:
This adds a new resource for associating instance IDs, which will only
need to be used in edge-cases. We'll auto-detect instance identifiers
for the most popular cases, and perform zero-trust authentication.

It seemed weird to separate the agent script and agent itself. This also
led to difficulties previewing the operating system prior to start.

[Before](https://github.com/coder/terraform-provider-coder/blob/549826ba3ae919dcddde79ce0d8442b35e1fe67c/examples/provider/provider.tf)
[After](https://github.com/coder/terraform-provider-coder/blob/45ff2c88e2a9a40d3872a499ade66a4d71ea578e/examples/provider/provider.tf)


internal/provider/provider.go
```diff
@@ -115,82 +144,90 @@ func New() *schema.Provider {
...
+						Type:        schema.TypeString,
+						Computed:    true,
+						Description: "Run this script on startup of an instance to initialize the agent.",
+					},
+					"arch": {
+						Type:         schema.TypeString,
+						ForceNew:     true,
+						Required:     true,
+						Description:  `The architecture the agent will run on. Must be one of: "amd64", "arm64".`,
+						ValidateFunc: validation.StringInSlice([]string{"amd64"}, false),
```
coadler, line 167: Validate doesn't include `arm64`

internal/provider/provider.go
```diff
@@ -115,82 +144,90 @@ func New() *schema.Provider {
...
-					rd.SetId(uuid.NewString())
-					err := rd.Set("token", uuid.NewString())
-					if err != nil {
-						return diag.FromErr(err)
-					}
+			"coder_agent_instance": {
+				Description: "Use this resource to associate an instance ID with an agent for zero-trust " +
+					"authentication. This association is done automatically for \"google_compute_instance\", " +
+					"\"aws_instance\", \"azurerm_linux_virtual_machine\", and " +
+					"\"azurerm_windows_virtual_machine\" resources.",
```
coadler, line 208: I find it slightly weird that we use the cloud names for aws and azure but not google. also, what does `azurerm` refer to?

coadler (approved):

internal/provider/provider.go
kylecarbs: Good catch

internal/provider/provider.go
kylecarbs: It's primarily because google doesn't in their resource naming, so I wanted to keep it consistent. `azurerm` stands for Azure Resource Manager- certainly _interestingly_ named...

coadler: Makes sense!

PR #31, opened 2022-07-18
Add data source attribute for owner email address

dwahler:
This PR adds a new attribute to the `coder_workspace` data source: `owner_email`, which is populated from the `CODER_WORKSPACE_OWNER_EMAIL` environment variable.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -76,12 +76,6 @@ func New() *schema.Provider {
 					}
 					_ = rd.Set("owner", owner)
 
-					ownerName := os.Getenv("CODER_WORKSPACE_OWNER_NAME")
-					if ownerName == "" {
-						ownerName = owner
-					}
-					_ = rd.Set("owner_name", ownerName)
-
 					ownerEmail := os.Getenv("CODER_WORKSPACE_OWNER_EMAIL")
 					_ = rd.Set("owner_email", ownerEmail)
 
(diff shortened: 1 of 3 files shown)


internal/provider/provider_test.go
```diff
@@ -19,7 +19,10 @@ func TestProvider(t *testing.T) {
 }
 
 func TestWorkspace(t *testing.T) {
-	t.Parallel()
+	t.Setenv("CODER_WORKSPACE_OWNER", "owner123")
+	t.Setenv("CODER_WORKSPACE_OWNER_NAME", "Workspace Owner Jr.")
```
kylecarbs, line 23: I kinda messed this up in the other issue, and I think it could be better to remove this for now. Since we don't have a display name, it could be easy to confuse `_name` with username.

kylecarbs (changes requested):

internal/provider/provider.go
```diff
@@ -69,31 +69,46 @@ func New() *schema.Provider {
...
 					_ = rd.Set("owner", owner)
+
+					ownerName := os.Getenv("CODER_WORKSPACE_OWNER_NAME")
+					if ownerName == "" {
+						ownerName = owner
+					}
+					_ = rd.Set("owner_name", ownerName)
+
+					ownerEmail := os.Getenv("CODER_WORKSPACE_OWNER_EMAIL")
+					_ = rd.Set("owner_email", ownerEmail)
```
dwahler, line 86: The other attributes handled by this function all have a non-empty default value in case the expected environment var is missing. But I can't think of a good default for the email address, and in my testing, there didn't seem to be any problem with leaving it empty.

internal/provider/provider_test.go
dwahler: Sure. I was thinking it would be convenient to have a separate variable, because in the future we'll probably be able to get a display name from e.g. an OIDC provider. But you're right that for now, it's potentially confusing.

kylecarbs: That's a good idea to expose the attributes. Interesting

kylecarbs (approved):
Pending `owner_name` is removed in favor of `owner`, this is good to merge! We should cut a new release too so it can be consumed by your other PR in coder/coder.

PR #34, opened 2022-07-27
feat: Add "coder_metadata" resource

dwahler:
This PR adds support for the `coder_metadata` resource, which allows attaching arbitrary key/value metadata to other resources in a template, as described in coder/coder#2989.


internal/provider/provider.go
```diff
@@ -318,6 +320,74 @@ func New() *schema.Provider {
...
+									Type: schema.TypeBool,
+									Description: "Set to \"true\" to for items such as API keys whose values should be " +
+										"hidden from view by default. Note that this does not prevent metadata from " +
+										"being retrieved using the API, so it is not suitable for secrets that should " +
+										"not be exposed to workspace users.",
+									ForceNew: true,
+									Optional: true,
+									Default:  false,
+								},
+								"is_null": {
```
kylecarbs, line 381: Would we just assume an empty string is null? Maybe that's too lossy?

dwahler: Possibly! I implemented it this way because the original ticket asked for null support, and I figured somebody who's used to Terraform's semantics wouldn't necessarily be expecting us to conflate null and "" the way Go does. But if we're OK with that behavior, I definitely wouldn't mind getting rid of this hack.

kylecarbs (commented):
- `make gen` should be used to generate docs
- I'm looking for alternatives to names to `pair`, but I have no ideas! Maybe @mafredri has some thoughts 👀

dwahler:
> `make gen` should be used to generate docs

Done! I added an example as well.

> I'm looking for alternatives to names to `pair`, but I have no ideas! Maybe @mafredri has some thoughts 👀

Hmm, a few other possibilities that come to mind: 
* item
* entry
* field
* element
* member
* datum

docs/resources/metadata.md
```diff
@@ -0,0 +1,61 @@
...
+  count = data.coder_workspace.me.start_count
+}
+
+resource "coder_metadata" "pod_info" {
+  count = data.coder_workspace.me.start_count
+  resource_id = kubernetes_pod.dev[0].id
+  pair {
+    key = "pod_uid"
+    value = kubernetes_pod.dev[0].uid
+  }
```
ammario, line 29: I was confused about why `key` and `value` were nested until @kylecarbs explained to me that HCL allows multiple pair blocks. We should demonstrate that here.

ammario (dismissed):

docs/resources/metadata.md
dwahler: Good point! I updated the example to use multiple blocks, and also to demonstrate the `sensitive` attribute.

mafredri:
> I'm looking for alternatives to names to pair, but I have no ideas! Maybe @mafredri has some thoughts 👀

I do like `entry` (and `item` as a close second) proposed by @dwahler.

Some other options:
- `kv`
- `keyvalue`
- `group` (kind of like `pair`, though)

And some more "out there" options:
- `attach`
- `add`

I'd read these as "attach/add key value to coder metadata".

ammario:
`item` makes most sense to me, but I also find `pair` fine.

kylecarbs (approved):
Let's rename `pair` to `item` and it LGTM! We'll do a release immediately after too :)

PR #36, opened 2022-08-09
feat: Add `coder_provisioner` resource for attributes from the host

kylecarbs:
Our [Docker template](https://github.com/coder/coder/blob/main/examples/templates/docker/main.tf#L16-L60) exposes variables
which are easily obtainable from the host. By adding this resource, we allow simple setups that depend on host
architecture to work without input from the user.


bpmct:
Fixes coder/coder#1879

internal/provider/provider.go
```diff
@@ -150,6 +151,28 @@ func New() *schema.Provider {
 					},
 				},
 			},
+			"coder_host": {
```
bpmct, line 154: nit: I'm wondering now if `coder_host` is the best term for this because this technically checks the provisioner's environment, right? 

By default, the Coder host, provisioner, (and docker socket) all run in the same environment, so the outcome is still positive for Docker template users.

Deployments that eventually add eventual provisioners (and/or docker hosts) will understand the separation of concerns so they won't really need to dynamically fetch the host's environment, but there _may_ be some merit in understanding which architecture the provisioner is on? (which this does, but it calls it the host). Still an obscure use case and it's really meant to help with single-machine deployments.

I don't really have a problem with it being called coder_host, just thinking out loud and maybe we can link to this comment in the future if it comes up again.

internal/provider/provider.go
```diff
@@ -150,6 +151,28 @@ func New() *schema.Provider {
...
+				Description: "Use this data source to get information about the host.",
```
bpmct, line 155:
```suggestion
				Description: "Use this data source to get information about the host running Coder.",
```

internal/provider/provider_test.go
```diff
@@ -52,6 +53,33 @@ func TestWorkspace(t *testing.T) {
...
+		IsUnitTest: true,
+		Steps: []resource.TestStep{{
+			Config: `
+			provider "coder" {
+			}
+			data "coder_host" "me" {
+			}`,
+			Check: func(state *terraform.State) error {
+				require.Len(t, state.Modules, 1)
+				require.Len(t, state.Modules[0].Resources, 1)
```
bpmct, line 70: is this just ensuring properties exist? when would this fail?

internal/provider/provider.go
```diff
@@ -150,6 +151,28 @@ func New() *schema.Provider {
...
+				ReadContext: func(c context.Context, rd *schema.ResourceData, i interface{}) diag.Diagnostics {
+					rd.SetId(uuid.NewString())
+					rd.Set("os", runtime.GOOS)
+					rd.Set("arch", runtime.GOARCH)
+
+					return nil
+				},
+				Schema: map[string]*schema.Schema{
+					"os": {
```
bpmct, line 164: nit: Is it worth limiting the values to https://github.com/golang/go/blob/master/src/go/build/syslist.go or linking to the known values in the docs so users don't have to do guesswork? 

🤷🏼 I'm not sure if anyone would use this beyond what we're doing in the Docker template though.

bpmct (approved):
LGTM.

internal/provider/provider.go
kylecarbs: I think it is the host of the provisioner though, or at least that's what I'm intending for it to portray. 

We could alternatively do `coder_provisioner_host` or `coder_provisioner`. Maybe `coder_provisioner` would be better, because it seems likely we'll expose other options from that eventually :thinking:. Thoughts?

internal/provider/provider_test.go
kylecarbs: This just ensures it exists, otherwise it would panic below. Not entirely necessary for this test, just makes the output a bit nicer in case of a dev mistake!

internal/provider/provider.go
kylecarbs: I'll mention that it's the same as `GOOS`, and link to https://pkg.go.dev/runtime#pkg-constants.

internal/provider/provider.go
kylecarbs: I changed it to `coder_provisioner`. It makes more sense IMO.

internal/provider/provider_test.go
bpmct: 👍🏼

PR #37, opened 2022-08-15
feat: Expose access_port on coder_workspace to allow for alternative hosts

kylecarbs:
This allows our Docker template to use a custom host with the correct port.

### DIFF
--- a/docs/data-sources/workspace.md
+++ b/docs/data-sources/workspace.md
@@ -26,6 +26,7 @@ resource "kubernetes_pod" "dev" {
 
 ### Read-Only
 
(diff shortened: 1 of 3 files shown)


bpmct (commented):
Does `coder server` always assume HTTP or HTTPS as an access URL now?  Should we add an access_port there too?

kylecarbs:
Nah, this just pulls the access port dynamically from the URL itself. I think it's better to be in the URL, this just allows for situations where you want a custom host, but same port.

bpmct (approved):

PR #43, opened 2022-08-18
Fix access url when host is not provided

spikecurtis:
Fixes empty host b0rking the URL.


kylecarbs (approved):

kylecarbs:
#44 merges this in, with the original code too.

PR #48, opened 2022-09-09
feat: Add icon to coder_metadata resource

BrunoQuaresma:
Related to https://github.com/coder/coder/issues/3487

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -406,6 +406,12 @@ func New() *schema.Provider {
 						ForceNew:    true,
 						Optional:    true,
 					},
+					"icon": {
+						Type:        schema.TypeString,
+						Description: "The Material UI icon name in CamelCase.",
+						ForceNew:    true,
+						Optional:    true,
+					},
 					"item": {
 						Type:        schema.TypeList,
 						Description: "Each \"item\" block defines a single metadata item consisting of a key/value pair.",
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -407,10 +407,19 @@ func New() *schema.Provider {
 						Optional:    true,
 					},
 					"icon": {
-						Type:        schema.TypeString,
-						Description: "The Material UI icon name in CamelCase.",
-						ForceNew:    true,
-						Optional:    true,
+						Type: schema.TypeString,
+						Description: "A URL to an icon that will display in the dashboard. View built-in " +
+							"icons here: https://github.com/coder/coder/tree/main/site/static/icons. Use a " +
+							"built-in icon with `data.coder_workspace.me.access_url + \"/icons/<path>\"`.",
+						ForceNew: true,
+						Optional: true,
+						ValidateFunc: func(i interface{}, s string) ([]string, []error) {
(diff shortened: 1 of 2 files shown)


BrunoQuaresma:
If you approve this and have permission to merge, please do it 🚀

internal/provider/provider.go
```diff
@@ -406,6 +406,12 @@ func New() *schema.Provider {
 						ForceNew:    true,
 						Optional:    true,
 					},
+					"icon": {
+						Type:        schema.TypeString,
+						Description: "The Material UI icon name in CamelCase.",
```
deansheather, line 411: Instead of letting people pick any icon we should just provide a list of icon types such as `["compute", "storage", "unknown"]` etc. and map those to icons. This way we avoid locking ourselves into material UI icons

BrunoQuaresma: Hm... I think this is reasonable but kinda limited. No strong opinions tho. cc: @bpmct

bpmct: Hmm... kinda sucks that we can't do the entire set but I understand why as we might want to change icon sets down the road.

Thoughts on also supporting arbitrary icon URLs, the same way the `coder_app` resource does? This shouldn't necessarily be blocking, but ideally these things should be as customizable as possible as long as there aren't technical limitations... For example, in templates:

- use any resource
- use any name
- use any parameters
- add any app, with any icon URL
- add any metadata
- add any instructions in README
- ideally: use any icon for resources

bpmct (approved):

internal/provider/provider.go
deansheather: In the interest of keeping things simple since this is not a very important feature IMO unlike app icons, we should do the URL support in the future

kylecarbs: I think we should only do icon URLs, not `material-ui` icons. We don't use the latest `material-ui` version, so it might be difficult for users to understand what they can or cannot use.

The relative path has been working well for us, so I'd say we go down that route. Thoughts @bpmct ?

bpmct: I do like using a common "image" mechanism for templates, apps, and resources.  Perhaps we could download some svg/material icons and serve them statically?

BrunoQuaresma:
After seeing @kylecarbs suggestion and talking with @bpmct I think having this icon the same as app icons is the best so it supports Icon URLs and we will upload some built-in icons.

internal/provider/provider.go
```diff
@@ -406,6 +406,21 @@ func New() *schema.Provider {
 						ForceNew:    true,
 						Optional:    true,
 					},
+					"icon": {
+						Type: schema.TypeString,
+						Description: "A URL to an icon that will display in the dashboard. View built-in " +
+							"icons here: https://github.com/coder/coder/tree/main/site/static/icons. Use a " +
```
kylecarbs, line 412:
```suggestion
							"icons here: https://github.com/coder/coder/tree/main/site/static/icon. Use a " +
```

kylecarbs (approved):

deansheather (approved):

PR #49, opened 2022-09-12
fix: Add check for duplicate metadata keys

kylecarbs:
Fixes coder/coder#3721.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -539,7 +539,8 @@ func populateIsNull(resourceData *schema.ResourceData) (result interface{}, err 
 	return resultItems, nil
 }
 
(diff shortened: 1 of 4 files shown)


internal/provider/provider.go
```diff
@@ -534,7 +541,7 @@ func populateIsNull(resourceData *schema.ResourceData) (result interface{}, err
 
 // valueAsString takes a cty.Value that may be a string or null, and converts it to either a Go string
```
dwahler, line 542: This comment was wrong to begin with 😞 Might as well fix it now:

```suggestion
// valueAsString takes a cty.Value that may be a string or null, and converts it to a Go string,
// which will be empty if the input value was null.
```

dwahler (approved):

PR #51, opened 2022-09-18
fix: Regenerate agent token on each run

kylecarbs:
This resolves a race condition where an old agent could be updated and connected to before the new agent was registered.


dwahler (approved):
I can't say I entirely understand what race condition this fixes, but the change itself seems fine to me.

kylecarbs:
@dwahler I should shed some light on it anyways!

An agent could connect to an old workspace agent if started twice. This can happen if provisioning and starting completes before the agent token is returned from a provisioner daemon. With a unique auth token, this can't happen anymore!

PR #52, opened 2022-09-21
feat: add healthcheck parameters to coder_app

f0ssel:
Blocking https://github.com/coder/coder/pull/4114


internal/provider/provider.go
```diff
@@ -373,6 +373,34 @@ func New() *schema.Provider {
...
 						ConflictsWith: []string{"command"},
 					},
+					"healthcheck_enabled": {
+						Type:          schema.TypeBool,
+						Description:   "Run a HTTP request periodically to determine the application readiness.",
+						ForceNew:      true,
+						Optional:      true,
+						ConflictsWith: []string{"command"},
+					},
+					"healthcheck_url": {
```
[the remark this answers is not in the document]
deansheather: The healthcheck probably won't be the root path in some cases.

internal/provider/provider.go
```diff
@@ -373,6 +373,34 @@ func New() *schema.Provider {
...
+						Type:          schema.TypeString,
+						Description:   "HTTP address used determine the application readiness.",
```
deansheather, line 385: Should document what constitutes a pass or fail. "Any 2xx response within the 5 second timeout window is considered a pass" or something.

Also add this sentence to the app.md file too.

internal/provider/provider.go
```diff
@@ -373,6 +373,34 @@ func New() *schema.Provider {
 						Optional:      true,
 						ConflictsWith: []string{"command"},
 					},
+					"healthcheck_enabled": {
+						Type:          schema.TypeBool,
+						Description:   "Run a HTTP request periodically to determine the application readiness.",
+						ForceNew:      true,
+						Optional:      true,
+						ConflictsWith: []string{"command"},
+					},
```
deansheather, line 382: Can `healthcheck_enabled` be implied by the existence of `healthcheck_url` instead of having it's own value?

deansheather (approved):

internal/provider/provider.go
f0ssel: I'm down for that

internal/provider/provider.go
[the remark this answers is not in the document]
f0ssel: Yeah the root could be a big webapp, when we really just need a 200.

examples/resources/coder_app/resource.tf
```diff
@@ -11,11 +11,14 @@ EOF
...
-  url               = "http://localhost:13337"
-  relative_path     = true
+  agent_id      = coder_agent.dev.id
+  name          = "VS Code"
+  icon          = data.coder_workspace.me.access_url + "/icons/vscode.svg"
+  url           = "http://localhost:13337"
+  relative_path = true
+  healthcheck_url = "http://localhost:13337/healthz"
+  healthcheck_interval = 5
+  healthcheck_threshold = 6
```
kylecarbs, line 21: Should we put these inside a `healthcheck` block instead?
```
healthcheck {
  url = "http://..." 
  interval = 5
  threshold = 6
}
```

internal/provider/provider.go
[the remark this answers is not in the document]
kylecarbs: Yup, that makes a lot of sense.

docs/resources/app.md
```diff
@@ -31,6 +31,9 @@ resource "coder_app" "code-server" {
   icon          = data.coder_workspace.me.access_url + "/icons/vscode.svg"
   url           = "http://localhost:13337"
   relative_path = true
+  healthcheck_url = "http://localhost:13337/healthz"
+  healthcheck_interval = 5
+  healthcheck_threshold = 6
```
deansheather, line 36: reformat probably

examples/resources/coder_app/resource.tf
```diff
@@ -16,6 +16,9 @@ resource "coder_app" "code-server" {
   icon          = data.coder_workspace.me.access_url + "/icons/vscode.svg"
   url           = "http://localhost:13337"
   relative_path = true
+  healthcheck_url = "http://localhost:13337/healthz"
+  healthcheck_interval = 5
+  healthcheck_threshold = 6
```
deansheather, line 21: reformat

internal/provider/provider.go
```diff
@@ -373,33 +373,29 @@ func New() *schema.Provider {
...
-						Type:          schema.TypeBool,
-						Description:   "Run a HTTP request periodically to determine the application readiness.",
-						ForceNew:      true,
-						Optional:      true,
-						ConflictsWith: []string{"command"},
-					},
 					"healthcheck_url": {
 						Type:          schema.TypeString,
-						Description:   "HTTP address used determine the application readiness.",
+						Description:   "HTTP address used determine the application readiness. " + "A successful health check is a HTTP response code less than 500 returned before healthcheck_interval seconds.",
```
deansheather, line 378: Why the string concatenation?

deansheather (approved):

examples/resources/coder_app/resource.tf
f0ssel: I considered this but ultimately didn't because I didn't want to make a new codersdk type for it on the api side and wanted the API and the terraform to share the same schema if possible. If we are cool with the TF being different from the API type schema I'm down for just the TF side.

kylecarbs: Why not create a new type in `codersdk` for it? It seems kinda like a separate entity, and it makes more sense from my POV for this to be a block.

kylecarbs: It also cleanly separates that health checking is disabled if the block isn't specified, which is kinda nice.

f0ssel: Sure I'm down if we like that

kylecarbs: I'm personally a fan, and I believe @deansheather is too based on the :+1: :eyes: 

shiiip it

bpmct (commented):
Huge fan of the schema. Seems like its exactly what we're looking for

kylecarbs (approved):

internal/provider/provider.go
```diff
@@ -373,6 +373,35 @@ func New() *schema.Provider {
...
+									Description: "Duration in seconds to wait between healthcheck requests.",
+									ForceNew:    true,
+									Required:    true,
+								},
+								"threshold": {
+									Type:        schema.TypeInt,
+									Description: "Number of consecutive heathcheck failures before returning an unhealthy status.",
+									ForceNew:    true,
+									Required:    true,
+								},
```
kylecarbs, line 401: One minor nit before merge, should we make these optional instead with reasonable defaults?

Also Kubernetes calls this `period` instead of `interval`. I'm impartial, just wanted to toss the idea up.

kylecarbs (approved):

PR #59, opened 2022-10-03
fix: Add error type to parameters

kylecarbs:


docs/resources/app.md
```diff
@@ -41,13 +41,13 @@ resource "coder_app" "code-server" {
 resource "coder_app" "vim" {
   agent_id = coder_agent.dev.id
   name     = "Vim"
-  icon     = "${data.coder_workspace.me.access_url}/icon/vim.svg"
+  icon     = data.coder_workspace.me.access_url + "/icons/vim.svg"
```
bpmct, line 44: This actually doesn't work

```sh
  Error: Invalid operand
  Unsuitable value for right operand: a number is required.
✔ Detecting persistent resources [4940ms]
✘ Cleaning Up [110ms]
template import provision for start: recv import provision: plan terraform: terraform plan: exit status 1
Run 'coder templates push --help' for usage.  
```

provider/parameter_test.go
```diff
@@ -73,10 +73,24 @@ data "coder_parameter" "region" {
...
+data "coder_parameter" "region" {
+	name = "Region"
+	type = "string"
+	default = "hello"
+	validation {
+		regex = "hello"
+	}
+}
+`,
+		ExpectError: regexp.MustCompile("an error must be specified"),
```
bpmct, line 93: This seems reasonable given Terraform expects an error message for all validation conditions too.

bpmct (changes requested):

docs/resources/app.md
kylecarbs: Fixed

PR #60, opened 2022-10-04
feat: Allow for specifying a custom host for the access URL

kylecarbs:
This should fix our Docker woes...


deansheather (commented):
doesn't work if the port doesn't match i.e. if you use `--tunnel` the port will be 443 but it should be 8080 or whatever

kylecarbs:
Yup, we'll pause on this

PR #62, opened 2022-10-06
feat: add app share property

deansheather:
Add `share` property to workspace app resources of type string. Must be one of `"owner"`, `"template"`, `"authenticated"`, `"public"`.


docs/resources/app.md
```diff
@@ -26,11 +26,12 @@ EOF
...
-  agent_id  = coder_agent.dev.id
-  name      = "VS Code"
-  icon      = data.coder_workspace.me.access_url + "/icons/vscode.svg"
-  url       = "http://localhost:13337"
-  subdomain = false
+  agent_id      = coder_agent.dev.id
+  name          = "VS Code"
+  icon          = data.coder_workspace.me.access_url + "/icons/vscode.svg"
+  url           = "http://localhost:13337"
+  sharing_level = "owner"
```
kylecarbs, line 33: What do you think about `access` as a name instead?

deansheather: so just `access = "owner"`?

deansheather: after talking in slack we decided on `share` instead

provider/app.go
```diff
@@ -76,6 +77,39 @@ func appResource() *schema.Resource {
 				ForceNew: true,
 				Optional: true,
 			},
+			"share": {
+				Type: schema.TypeString,
+				Description: "Determines the sharing level of the app. " +
```
kylecarbs, line 82: I'd get rid of this first sentence. It's pretty implicit.

provider/app.go
```diff
@@ -76,6 +77,39 @@ func appResource() *schema.Resource {
...
+					`Valid values are "owner", "template", "authenticated" ` +
+					`and "public". Level "owner" disables sharing on the ` +
+					"app, so only the workspace owner can access it. Level " +
+					`"template" shares the app with all users that can read ` +
+					`the workspace's template. Level "authenticated" shares ` +
+					`the app with all authenticated users. Level "public" ` +
+					"shares it with any user, including unauthenticated " +
+					"users. Permitted application sharing levels can be " +
+					`controlled via a flag on "coder server". Defaults to ` +
+					`"owner" (sharing disabled).`,
```
kylecarbs, line 95: We should wait to allow the blocking of sharing globally IMO.

kylecarbs (approved):

provider/app.go
deansheather: Talked to Kyle on slack and he agrees we should keep the flag but it should default to all levels being available if you're entitled to the feature.

PR #63, opened 2022-10-07
chore: fix app share parameter description

deansheather:


bpmct (approved):
We should capitalize enterprise but LGTM

PR #68, opened 2022-10-18
feat: add slug to app, rename name to display_name

deansheather:
- Adds new required `slug` property to `coder_app`
    - Must match a regex that ensures the slug is hostname-safe
- Adds new property `display_name` to `coder_app`
- Deprecates `name` in favor of `display_name` in `coder_app`

coder/coder#4573

### DIFF
--- a/provider/app.go
+++ b/provider/app.go
@@ -3,11 +3,24 @@ package provider
 import (
 	"context"
 	"net/url"
+	"regexp"
 
 	"github.com/google/uuid"
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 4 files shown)


provider/app.go
```diff
@@ -77,6 +111,15 @@ func appResource() *schema.Resource {
 				ForceNew: true,
 				Optional: true,
 			},
+			"relative_path": {
```
kylecarbs, line 114: We should make sure this gets removed in all of our templates then!

kylecarbs (approved):

provider/app.go
deansheather: I just moved this value to be below subdomain, it's already been deprecated for a while now. I deprecated it a few weeks ago when we added the subdomain key.

PR #74, opened 2022-11-10
Add cost field to resource_metadata

ammario:
Iffy on whether to call it `quota_cost` or `cost`. I'm leaning `cost` because it's shorter, and we may use these fields to show analytics even when quotas are disabled.


kylecarbs (commented):
Run `make gen` before merge!

kylecarbs (approved):

PR #78, opened 2022-11-22
feat: Add motd_file to coder_agent

mafredri:
This change allows us to enable better feature parity with OpenSSH by
being able to display a message when users login. This is useful for
template authors who want to inform their users about changes or
reporting about the state of the workspace.

I did consider if we should turn this into a script, like:

```tf
motd_script = <<-EOS
	cat /etc/motd
EOS
```

But while this is flexible, it raises questions like how often should we run it, sould we run it on every query, what if it takes a long time to execute, etc. For this reason we're sticking with the simplicity of a single file and making it the responsibility of the author to start a service to update its content, where required.

This feature could be expanded to support things like VSCode/JetBrains extensions as well, to surface the message not just via SSH, but these clients as well.

### DIFF
--- a/provider/agent.go
+++ b/provider/agent.go
@@ -101,6 +101,12 @@ func agentResource() *schema.Resource {
 				Optional:    true,
 				Description: "A URL to a document with instructions for troubleshooting problems with the agent.",
 			},
+			"motd_file": {
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Optional:    true,
+				Description: "The path to a file within the workspace containing a message to display to users when they login via SSH. A typical value would be /etc/motd.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


provider/agent.go
```diff
@@ -101,6 +101,12 @@ func agentResource() *schema.Resource {
 				Optional:    true,
 				Description: "A URL to a document with instructions for troubleshooting problems with the agent.",
 			},
+			"motd_file": {
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Optional:    true,
+				Description: "The path to a file within the workspace containing a message to display to users when they login via SSH. A typical value would be /etc/motd.",
```
kylecarbs, line 108: Should we default to /etc/motd?

kylecarbs (approved):

provider/agent.go
mafredri: It's a good question. I was thinking about the following:
- The path will be nonsensical on Windows, so the default would apply to Unix only
- Is it weird that it's on-by-default?

For instance, with on-by-default. Users using `alpine` as their workspace base would be greeted by:

```
Welcome to Alpine!

The Alpine Wiki contains a large amount of how-to guides and general
information about administrating Alpine systems.
See <http://wiki.alpinelinux.org/>.

You can setup the system with the command: setup-alpine

You may change this message by editing /etc/motd.
```

It's not necessarily a bad thing though, I'm definitely open to defaulting to `/etc/motd`.

kylecarbs: Let's leave it out for now, and if we see fit we can always add a default!

PR #79, opened 2022-11-24
feat: Add support for shutdown_script

mtojek:
Issue: https://github.com/coder/coder/issues/4677

This PR introduces a new parameter - the shutdown script. The script is intended to be executed before the agent is stopped, before the [networking is shut down](https://github.com/coder/coder/blob/main/agent/agent.go#L930).


mafredri (approved):
Do you think we will need a shutdown_script_timeout as well? I think it'd be fine to wait indefinitely until shutdown script has completed, at least to start with. Most likely someone will want to debug something if the shutdown script has hung.

mtojek:
Yes, the idea is to come back quickly with basic feature. Then, I can extend it with extras like configurable execution timeout, retries, etc.

johnstcn (approved):

mtojek:
Hey @kylecarbs, it seems that I'm not authorized to merge this PR. Could you please help with this?

PR #84, opened 2023-01-18
feat: Add `startup_script_timeout` and `delay_login_until_ready`

mafredri:
Ref: https://github.com/coder/coder/issues/5749

### DIFF
--- a/docs/resources/agent.md
+++ b/docs/resources/agent.md
@@ -46,13 +46,15 @@ resource "kubernetes_pod" "dev" {
 
 ### Optional
 
+- `allow_login_before_ready` (Boolean) Allow users to login to the workspace before the agent is ready. Note that, when enabled, the agent could still be executing the startup script and the workspace in an incomplete state when logging in.
 - `auth` (String) The authentication type the agent will use. Must be one of: "token", "google-instance-identity", "aws-instance-identity", "azure-instance-identity".
 - `connection_timeout` (Number) Time in seconds until the agent is marked as timed out when a connection with the server cannot be established. A value of zero never marks the agent as timed out.
 - `dir` (String) The starting directory when a user creates a shell session. Defaults to $HOME.
 - `env` (Map of String) A mapping of environment variables to set inside the workspace.
 - `motd_file` (String) The path to a file within the workspace containing a message to display to users when they login via SSH. A typical value would be /etc/motd.
-- `shutdown_script` (String) A script to run before the agent is stopped.
-- `startup_script` (String) A script to run after the agent starts.
+- `shutdown_script` (String) A script to run before the agent is stopped. The script should exit when it is done to signal that the workspace can be stopped.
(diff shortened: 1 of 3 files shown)


provider/agent.go
```diff
@@ -77,10 +77,24 @@ func agentResource() *schema.Resource {
...
+				ForceNew:     true,
+				Optional:     true,
+				Description:  "Time in seconds until the agent ready status is marked as timed out, this happens when the startup script has not completed (exited) in the given time.",
+				ValidateFunc: validation.IntAtLeast(1),
+			},
+			"shutdown_script": {
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Optional:    true,
+				Description: "A script to run before the agent is stopped. The script should exit when it is done to signal that the workspace can be stopped.",
```
mafredri, line 96: Updated description here, I considered commenting it out or adding `(NOT YET IMPLEMENTED.)`, but we'll likely do this soon after we get the startup script changes in.

docs/resources/agent.md
```diff
@@ -46,13 +46,15 @@ resource "kubernetes_pod" "dev" {
 
 ### Optional
 
+- `allow_login_before_ready` (Boolean) Allow users to login to the workspace before the agent is ready. Note that, when enabled, the agent could still be executing the startup script and the workspace in an incomplete state when logging in.
```
mafredri, line 49: @kylecarbs does this match what you were suggesting in https://github.com/coder/coder/issues/5749#issuecomment-1385877505?

I interpreted it as an option to allow/prevent it. Is that how you intended it? Or did you intend it to be the "default behavior for login before ready"? If it's the latter, we could rename this to be just `login_before_ready` instead of allow.

kylecarbs: I might invert the naming scheme because, eventually we'll make this default to blocking login before ready, and ideally it'd be false by default.

Apart from that we're aligned!

mtojek (approved):
LGTM, although it might be valuable to RFC/describe incoming changes. It will be definitely more complex than adding a single timeout.

mafredri:
> LGTM, although it might be valuable to RFC/describe incoming changes. It will be definitely more complex than adding a single timeout.

Re: RFC, is there any specific questions you have after reading https://github.com/coder/coder/issues/5749? I felt it adequately describes what will be implemented but I'm happy to answer any questions. IMO it won't be a very complex change since it's focused towards fixing the `startup_script` use-case only, leaving room for extending it to support `shutdown_script`.

mtojek:
Yes, I have a feeling that we're missing a short paragraph mainly describing changes: `WorkspaceAgentStatus` vs `WorkspaceAgentState`, deprecation path for these properties.

It's hard to figure out from `Introduce new WorkspaceAgentStatus: starting, start_timeout, started, (stopping, stop_timeout, stopped)`, what's the intention and outcomes without jumping about pages. I believe that users can find answers in the comments, but a single write-up would be appreciated.

PR #90, opened 2023-02-03
feat: Validate monotonicity for coder_parameter

mtojek:
Related: https://github.com/coder/coder/issues/5574
Related: https://github.com/coder/coder/issues/5978

This PR adds a new validation property `monotonic` to the `coder_parameter`. Unfortunately, we can only validate the format within the Terraform runner, but a comparison of values, current and new, will be performed on the `coderd` API level.

### DIFF
--- a/provider/parameter.go
+++ b/provider/parameter.go
@@ -25,11 +25,18 @@ type Option struct {
 }
 
 type Validation struct {
-	Min   int
-	Max   int
+	Min       int
+	Max       int
+	Monotonic string
+
 	Regex string
 	Error string
 }
+
+const (
+	ValidationMonotonicIncreasing = "increasing"
+	ValidationMonotonicDecreasing = "decreasing"
+)
 
 type Parameter struct {
 	Value       string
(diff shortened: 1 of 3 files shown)


provider/parameter.go
```diff
@@ -235,9 +242,14 @@ func parameterDataSource() *schema.Resource {
 							Description:  "The maximum of a number parameter.",
 							RequiredWith: []string{"validation.0.min"},
 						},
+						"monotonic": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Description: "Number monotonicity, either increasing or decreasing.",
```
mafredri, line 248: Just curious: Why don't we use a validation func here?

(I do see the check in `valueIsType`, it's just not obvious why we're doing that instead of validation func.)

mafredri (approved):

provider/parameter.go
mtojek: I will pass this question to @kylecarbs who originally implemented this code. I deduce that this is due to the compliance with `terraform-plugin-sdk`?

deansheather (approved):

provider/parameter.go
kylecarbs: Not sure actually...

kylecarbs (approved):

PR #94, opened 2023-02-07
test: parameter description contains Markdown

mtojek:
Related: https://github.com/coder/coder/issues/5931

This PR improves tests to present that the Markdown format is acceptable for the parameter description.

~This PR introduces a new Markdown property to render rich descriptions, with Markdown tags included (see example).~

~Comment:
I admit that I wanted to use the `description` and then strip Markdown tags from descriptions in CLI, but when I looked at the [logic](https://github.com/stiang/remove-markdown/blob/main/index.js) required to perform this (rewritten to Go), I decided to KISS and use a dedicated property, that is only processed by the site.~

### DIFF
--- a/examples/resources/coder_parameter/resource.tf
+++ b/examples/resources/coder_parameter/resource.tf
@@ -17,9 +17,8 @@ data "coder_parameter" "example" {
 
 data "coder_parameter" "ami" {
   name = "Machine Image"
-  description = "Provide the machine image."
-  description_markdown = <<-EOT
-    # Select the machine image
+  description = <<-EOT
+    # Provide the machine image
     See the [registry](https://container.registry.blah/namespace) for options.
     EOT
   option {
--- a/examples/resources/coder_parameter/resource.tf
+++ b/examples/resources/coder_parameter/resource.tf
@@ -17,6 +17,11 @@ data "coder_parameter" "example" {
 
 data "coder_parameter" "ami" {
   name = "Machine Image"
+  description = "Provide the machine image."
+  description_markdown = <<EOT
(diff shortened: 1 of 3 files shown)


mafredri (approved):
LGTM!

kylecarbs (commented):
I'm leaning towards optimizing syntax for the web interface rather than the CLI, because it's presently unclear in the syntax when `description` would be shown and when `description_markdown ` would be.

What do you think about making `description` markdown by default, then add a `description_plaintext` for the CLI? Another option is to display the markdown as text in the CLI, which I don't think is bad. Markdown is pretty readable anyways (I think this might be my preference actually).

mtojek:
Thanks for jumping in, @kylecarbs.

> I'm leaning towards optimizing syntax for the web interface rather than the CLI, because it's presently unclear in the syntax when description would be shown and when description_markdown  would be.

TIL! I think that we have to highlight who is the more important consumer, CLI or site. We can use it as the rule of thumb that in our case it's always the site.

> What do you think about making description markdown by default, then add a description_plaintext for the CLI?

If we're site-oriented, then `description` should be kept as is, even if it uses Markdown tags. Additionally, we can expose an extra generated field in the API that shows the rendered description,  `description_plaintext`. I don't think that this is something we need to keep in the database, it can be rendered dynamically 👍 .

Let me know your thoughts.

kylecarbs:
That makes sense to me. I like it!

mtojek:
Maybe I will update examples to show that Markdown description is fine.

mafredri (approved):

PR #96, opened 2023-02-10
feat: Add `shutdown_script_timeout`

mafredri:
Ref: https://github.com/coder/coder/pull/6139

### DIFF
--- a/docs/data-sources/parameter.md
+++ b/docs/data-sources/parameter.md
@@ -56,6 +56,7 @@ Optional:
 - `error` (String) An error message to display if the value doesn't match the provided regex.
 - `max` (Number) The maximum of a number parameter.
 - `min` (Number) The minimum of a number parameter.
+- `monotonic` (String) Number monotonicity, either increasing or decreasing.
 - `regex` (String) A regex for the input parameter to match against.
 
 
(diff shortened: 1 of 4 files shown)


docs/data-sources/parameter.md
```diff
@@ -56,6 +56,7 @@ Optional:
 - `error` (String) An error message to display if the value doesn't match the provided regex.
 - `max` (Number) The maximum of a number parameter.
 - `min` (Number) The minimum of a number parameter.
+- `monotonic` (String) Number monotonicity, either increasing or decreasing.
```
mtojek, line 59: 👍 

nit: we may want to improve the Makefile to detect discrepancies.

mtojek (approved):
It's complimentary to the startup timeout, so 👍 .

docs/data-sources/parameter.md
mafredri: This is done via `make gen`, we should probably verify in CI that there are no changes.

mtojek: Yes, that's what I meant.

mafredri: I'll toss up a PR 👍🏻

PR #97, opened 2023-02-10
feat: Add flag to enable managed variables

mtojek:
Related: https://github.com/coder/coder/issues/5980

This PR introduces a feature flag to enable managed template-wide variables. The flag is required to select the processing strategy - either use legacy Parameter Schema or new template variables ([under development](https://github.com/coder/coder/pull/6134)).

Comment:
I had to add this flag as `terraform` complains:

```
> Upload "mtojek/docker-rich"? (yes/no) yes
✔ Queued [571ms]
✔ Setting up [1ms]
✔ Parsing template parameters [60ms]
⧗  Detecting persistent resources
  Terraform 1.2.6
  Error: Unsupported argument
  An argument named "feature_use_managed_variables" is not expected here.
✔ Detecting persistent resources [7623ms]
✘ Cleaning Up [110ms]
```

The idea is to remove the flag as the new strategy will be enabled by default.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -62,4 +62,5 @@ resource "google_compute_instance" "dev" {
 
 ### Optional
 
+- `feature_use_managed_variables` (Boolean) Feature: use managed Terraform variables. If disabled, Terraform variables will be included in legacy Parameter Schema.
 - `url` (String) The URL to access Coder.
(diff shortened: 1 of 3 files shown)


mafredri (approved):
We probably shouldn't remove the flag once we make it default, but we can mark it as deprecated.

mtojek:
As an alternative option, I was considering a `template_format_version`, but it might be too wide, and in general feature flags are useful. I don't have strong feelings about it, and I'm happy to change it if somebody has a good idea.

PR #107, opened 2023-03-02
feat: mark coder_parameter as optional

mtojek:
Related: https://github.com/coder/coder/issues/6100

This PR adds a new computed property to `coder_parameter`, `optional`.

Analogically to [Terraform variables](https://servian.dev/terraform-optional-variables-and-attributes-using-null-and-optional-flag-62c5cd88f9ca) - a parameter is considered to be required if it doesn't have a `default` property, or it's `= null`. An empty string is an acceptable value, and it means that the parameter is not required.

### DIFF
--- a/docs/data-sources/parameter.md
+++ b/docs/data-sources/parameter.md
@@ -32,7 +32,7 @@ Use this data source to configure editable options for workspaces.
 ### Read-Only
 
 - `id` (String) The ID of this resource.
-- `required` (Boolean) Whether this value is required.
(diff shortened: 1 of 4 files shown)


kylecarbs (approved):
Should we instead flip this to be optional? I feel like we implicitly might want options required.

mtojek:
Hey @kylecarbs, I flipped the property implementation to "optional". Feel free to re-review this changeset.

PR #109, opened 2023-03-07
feat: import value from legacy variable

mtojek:
Related: https://github.com/coder/coder/issues/6368

This PR introduces a new `coder_parameter` property, `legacy_variable`. The idea is to link it with the old parameter. See:

```terraform
provider "coder" {}

variable "old_account_name" {
  type    = string
  default = "fake-user" # for testing purposes, no need to set via env TF_...
}

data "coder_parameter" "account_name" {
  name                 = "Account Name"
  type                 = "string"
  legacy_variable_name = var.old_account_name
}
```

`coder_parameter.default` will be equal to the value of `var.old_account_name`, in this case `fake-user`.

Notice: once admin enables `feature_use_managed_variables = true`, users won't be able to the `legacy_variable` property. This is only for transition purposes. (old -> new params).

Comment:

I'm curious if it isn't good enough to directly link `default` with `var.old_account_name`:

```
data "coder_parameter" "account_name" {
  name                 = "Account Name"
  type                 = "string"
  default = var.old_account_name
}
```

^ @bpmct That would be a good experiment :)


kylecarbs (approved):
If we persist with the `var.xxx` syntax, I suppose migration would actually be seamless, because the prior value would be used :thinking:... that's a big :brain: idea!

bpmct (approved):

PR #111, opened 2023-03-14
feat: support list(string) as coder_parameter

mtojek:
Related: https://github.com/coder/coder/issues/6582

This PR adds another supported type to `coder_parameter`, a list of strings.

### DIFF
--- a/examples/resources/coder_parameter/resource.tf
+++ b/examples/resources/coder_parameter/resource.tf
@@ -75,3 +75,9 @@ data "coder_parameter" "fairy_tale" {
   name = "Fairy Tale"
   type = "string"
 }
+
+data "coder_parameter" "users" {
+  name = "System users"
+  type = "list(string)"
+  default = ["root", "user1", "user2"]
+}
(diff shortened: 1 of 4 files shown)


mafredri (approved):

examples/resources/coder_parameter/resource.tf
```diff
@@ -74,4 +74,10 @@ data "coder_parameter" "cat_lives" {
 data "coder_parameter" "fairy_tale" {
   name = "Fairy Tale"
   type = "string"
+}
+
+data "coder_parameter" "users" {
+  name    = "System users"
+  type    = "list(string)"
+  default = jsonencode(["root", "user1", "user2"])
```
mtojek, line 82: This is really unfortunate, but we need to pick a single type of property. We need to fallback to string.

BrunoQuaresma (approved):
LGTM.

examples/resources/coder_parameter/resource.tf
BrunoQuaresma: I think it is ok just wondering if `list:string` could be better to parse if we parse it at all. Just a thought, the current one looks good too.

mtojek: It would be definitely easier to parse, but it is aligned with Terraform now, so I'd rather keep like this.

PR #112, opened 2023-03-15
Add metadata block to the agent

ammario:


provider/agent_test.go
```diff
@@ -111,3 +111,49 @@ func TestAgentInstance(t *testing.T) {
...
+		Steps: []resource.TestStep{{
+			Config: `
+				provider "coder" {
+					url = "https://example.com"
+				}
+				resource "coder_agent" "dev" {
+					os = "linux"
+					arch = "amd64"
+					metadata {
+						key = "process_count"
```
kylecarbs, line 131: Thoughts on calling this `id` instead? It's consistent with other naming we've done in here.

kylecarbs (approved):

provider/agent_test.go
ammario: We can make it id... We don't really want people thinking they have to auto generate it though.

PR #120, opened 2023-05-10
feat: add session token data source

sreya:
fixes the TF side of https://github.com/coder/coder/issues/6354


mtojek (approved):
This is a similar change to [this](https://github.com/coder/terraform-provider-coder/commit/dd157523ed246d4018ce21bdadb8d9422586a63e), and I think that this is correct.

You can also drop full `.tf` file to examples to present the usage. There is `examples_test.go` which you can use to validate it.

PR #121, opened 2023-05-26
fix: coder_parameter validation min and max are truly optional

mtojek:
Related: https://github.com/coder/coder/issues/7582

WIP

### DIFF
--- a/provider/parameter.go
+++ b/provider/parameter.go
@@ -62,8 +62,25 @@ func parameterDataSource() *schema.Resource {
 		ReadContext: func(ctx context.Context, rd *schema.ResourceData, i interface{}) diag.Diagnostics {
 			rd.SetId(uuid.NewString())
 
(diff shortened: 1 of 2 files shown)


provider/parameter.go
```diff
@@ -325,6 +334,47 @@ func parameterDataSource() *schema.Resource {
...
+
+	// Fix the resource data
+	if rawValidationRule["min"].IsNull() {
+		validationRule["min"] = nil
+	}
+	if rawValidationRule["max"].IsNull() {
+		validationRule["max"] = nil
+	}
+
+	vArr[0] = validationRule
```
mafredri, line 374: Is this assignment needed? Since we're modifying a map it should be updated in place.

mafredri (approved):
Looks good 👍🏻

provider/parameter.go
mtojek: Fixed!

PR #125, opened 2023-06-02
Replace `login_before_ready` with `startup_script_behavior`

### LINKED ISSUE #124: Replace `login_before_ready` with `startup_script_behavior`
As part of https://github.com/coder/coder/issues/7757 we are renaming login before ready to clarify what this feature does.

- [ ] Deprecate `login_before_ready`
- [ ] Add `startup_script_behavior => blocking / non-blocking (default)`

mafredri:
Fixes #124

### DIFF
--- a/docs/resources/agent.md
+++ b/docs/resources/agent.md
@@ -50,12 +50,13 @@ resource "kubernetes_pod" "dev" {
 - `connection_timeout` (Number) Time in seconds until the agent is marked as timed out when a connection with the server cannot be established. A value of zero never marks the agent as timed out.
 - `dir` (String) The starting directory when a user creates a shell session. Defaults to $HOME.
 - `env` (Map of String) A mapping of environment variables to set inside the workspace.
-- `login_before_ready` (Boolean) This option defines whether or not the user can (by default) login to the workspace before it is ready. Ready means that e.g. the startup_script is done and has exited. When enabled, users may see an incomplete workspace when logging in.
+- `login_before_ready` (Boolean, Deprecated) This option defines whether or not the user can (by default) login to the workspace before it is ready. Ready means that e.g. the startup_script is done and has exited. When enabled, users may see an incomplete workspace when logging in.
 - `metadata` (Block List) Each "metadata" block defines a single item consisting of a key/value pair. This feature is in alpha and may break in future releases. (see [below for nested schema](#nestedblock--metadata))
 - `motd_file` (String) The path to a file within the workspace containing a message to display to users when they login via SSH. A typical value would be /etc/motd.
 - `shutdown_script` (String) A script to run before the agent is stopped. The script should exit when it is done to signal that the workspace can be stopped.
 - `shutdown_script_timeout` (Number) Time in seconds until the agent lifecycle status is marked as timed out during shutdown, this happens when the shutdown script has not completed (exited) in the given time.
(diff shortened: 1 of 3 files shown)


provider/agent_test.go
```diff
@@ -36,7 +36,7 @@ func TestAgent(t *testing.T) {
 					motd_file = "/etc/motd"
 					shutdown_script = "echo bye bye"
 					shutdown_script_timeout = 120
-					login_before_ready = false
+					startup_script_behavior = "blocking"
```
mtojek, line 39: nit: can we add some test cases when `startup_script_behavior` is non-blocking?

docs/resources/agent.md
```diff
@@ -50,12 +50,13 @@ resource "kubernetes_pod" "dev" {
...
 - `motd_file` (String) The path to a file within the workspace containing a message to display to users when they login via SSH. A typical value would be /etc/motd.
 - `shutdown_script` (String) A script to run before the agent is stopped. The script should exit when it is done to signal that the workspace can be stopped.
 - `shutdown_script_timeout` (Number) Time in seconds until the agent lifecycle status is marked as timed out during shutdown, this happens when the shutdown script has not completed (exited) in the given time.
 - `startup_script` (String) A script to run after the agent starts. The script should exit when it is done to signal that the agent is ready.
+- `startup_script_behavior` (String) This option sets the behavior of the startup_script. When set to "blocking", the startup_script must exit before the workspace is ready. When set to "non-blocking", the startup_script may run in the background and the workspace will be ready immediately. Default is "non-blocking", although "blocking" is recommended.
```
mtojek, line 59: nit: `startup_script` (code tag)

provider/agent.go
```diff
@@ -16,23 +16,31 @@ import (
 func agentResource() *schema.Resource {
 	return &schema.Resource{
 		Description: "Use this resource to associate an agent.",
-		CreateContext: func(c context.Context, resourceData *schema.ResourceData, i interface{}) diag.Diagnostics {
+		CreateContext: func(ctx context.Context, resourceData *schema.ResourceData, i interface{}) diag.Diagnostics {
```
mtojek, line 19: nit: I'm curious if there is a way to check if `CheckContext` called `updateStartupScriptBehaviorIfLoginBeforeReady` or if it was `ReadWithoutTimeout`.

BTW It would be nice to write a comment on why we need to call it in both contexts. I think that we usually stick with `ReadContext`.

mtojek, line 19: nit: I'm curious if there is a way to check if `CheckContext` called `updateStartupScriptBehaviorIfLoginBeforeReady` or if it was `ReadWithoutTimeout`.

BTW It would be nice to write a comment on why we need to call it in both contexts. I think that we usually stick with `ReadContext`.

mtojek (approved):
Nothing blocking 👍

provider/agent.go
mafredri: Both are called, but I never got this method working because the value reverts at the end resulting in `Step 1/1 error: After applying this test step, the plan was not empty.`.

This function was actually supposed to be removed in favor of empty default for `startup_script_behavior`, so it's good you asked.

provider/agent_test.go
mafredri: Added a bunch of test cases, although since there's no logic now it's a bit silly. 😄

mtojek (approved):

PR #137, opened 2023-07-04
feat: Prompt user for parameter values

mtojek:
Related: https://github.com/coder/coder/issues/6828
Blocked: https://github.com/coder/coder/issues/6828#issuecomment-1620044967 We discuss a bit different approach, where the user is not prompted but is able to set the transient parameter value via Parameters. This discussion must be sorted out before proceeding with this PR.

This PR adds a new property to the coder parameter to prompt the user for parameter values.

In the [original RFC](https://www.notion.so/coderhq/v2-Rich-parameters-re-prompt-a-value-between-builds-ac78d2366f054ea5af38fc01d6d017b5), we assumed only the `always` mode. It means that before the user starts the workspace, a prompt will show up with input fields to enter values. These parameter values will be included only in the current build (I will defer the auto-start behavior for later).

### DIFF
--- a/examples/resources/coder_parameter/resource.tf
+++ b/examples/resources/coder_parameter/resource.tf
@@ -74,8 +74,8 @@ data "coder_parameter" "cat_lives" {
 }
 
 data "coder_parameter" "fairy_tale" {
-  name = "Fairy Tale"
-  type = "string"
+  name        = "Fairy Tale"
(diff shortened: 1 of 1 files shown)


mtojek:
Closing based on https://github.com/coder/coder/issues/6828#issuecomment-1621831139

PR #138, opened 2023-07-05
feat: support ephemeral parameters

mtojek:
Related: https://github.com/coder/coder/issues/6828#issuecomment-1621831139

This PR adds new property for coder parameter - `ephemeral`. Values of ephemeral parameters are not persisted between workspace builds and can be just set for the current build.

### DIFF
--- a/provider/parameter.go
+++ b/provider/parameter.go
@@ -350,7 +350,7 @@ func parameterDataSource() *schema.Resource {
 				Type:        schema.TypeBool,
 				Default:     false,
 				Optional:    true,
-				Description: "The presence of the ephemeral dictates whether the parameter value will be preserved between consecutive workspace builds.",
+				Description: "The value of an ephemeral parameter will not be preserved between consecutive workspace builds.",
 			},
 			"legacy_variable_name": {
 				Type:         schema.TypeString,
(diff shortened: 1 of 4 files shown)


bpmct (approved):

provider/parameter.go
```diff
@@ -340,6 +346,12 @@ func parameterDataSource() *schema.Resource {
 				Optional:    true,
 				Description: "The order determines the position of a template parameter in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by name (ascending order).",
 			},
+			"ephemeral": {
+				Type:        schema.TypeBool,
+				Default:     false,
+				Optional:    true,
+				Description: "The presence of the ephemeral dictates whether the parameter value will be preserved between consecutive workspace builds.",
```
mafredri, line 353:
```suggestion
				Description: "The value of an ephemeral parameter will not be preserved between consecutive workspace builds.",
```

Small suggestion since I felt the original description left open what true/false means.

mafredri (approved):

provider/parameter.go
mtojek: Fixed

PR #140, opened 2023-07-06
feat: drop support for legacy variables

mtojek:
This PR drops support for legacy variables.

Since now, admins don't have to specify the flag `feature_use_managed_variables` as it will be enabled by default. The flag is marked as deprecated to not break the Terraform schema.


provider/provider.go
```diff
@@ -38,7 +38,9 @@ func New() *schema.Provider {
 			"feature_use_managed_variables": {
 				Type:        schema.TypeBool,
 				Description: "Feature: use managed Terraform variables. If disabled, Terraform variables will be included in legacy Parameter Schema.",
+				Default:     true,
```
mafredri, line 41: Future compatibility check: If we ever remove this from the provider, the default=true value won't be propagated and we'll essentially receive a false for this. Can this become a problem, will coder/coder handle it?

provider/provider.go
```diff
@@ -38,7 +38,9 @@ func New() *schema.Provider {
...
 				Optional:    true,
+				Deprecated:  "Terraform variables are now exclusively utilized for template-wide variables after the removal of support for legacy parameters.",
```
mafredri, line 43: This is deprecated, but still present. Should the description say something like: setting this is a no-op? Is that true or what would it mean for a template author to have this set to false?

mafredri (approved):

provider/provider.go
mtojek: Once this [PR](https://github.com/coder/coder/pull/8334) is merged, the flag will be ignored. Do you think that we should mention it in the schema description?

provider/provider.go
mtojek: Maybe we should add a sentence clarifying that this property is not used anymore?

provider/provider.go
mafredri: I think it may be worth a mention 👍🏻

mtojek: I added an extra sentence. Thanks for raising this!

PR #141, opened 2023-07-18
feat: ephemeral parameters must be optional

mtojek:
Related: https://github.com/coder/coder/issues/6828

This PR introduces another constraint for ephemeral parameters:
* ephemeral parameters can't be marked as required, so they must have the `default` property defined.

### DIFF
--- a/provider/parameter.go
+++ b/provider/parameter.go
@@ -134,6 +134,10 @@ func parameterDataSource() *schema.Resource {
 				return diag.Errorf("parameter can't be immutable and ephemeral")
 			}
 
+			if !parameter.Optional && parameter.Ephemeral {
+				return diag.Errorf("ephemeral parameter requires the default property")
+			}
+
 			if len(parameter.Validation) == 1 {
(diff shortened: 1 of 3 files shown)


provider/parameter.go
```diff
@@ -134,6 +134,10 @@ func parameterDataSource() *schema.Resource {
 				return diag.Errorf("parameter can't be immutable and ephemeral")
 			}
 
+			if !parameter.Optional && parameter.Ephemeral {
+				return diag.Errorf("ephemeral parameter requires the default property")
```
mafredri, line 138: Sanity check: do parameters have an optional field or is it inferred by setting default, and is that the only way to do it? (I always stop to ask when I see the check and message say something different.)

mafredri (approved):
LGTM!

provider/parameter.go
mtojek: > do parameters have an optional field

No.

> is it inferred by setting default

Yes 👍 

> is that the only way to do it

Yes 👍

PR #143, opened 2023-07-18
chore: bump go to 1.20 in release.yml

matifali:
I believe it was mistakenly left out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.17
+          go-version: 1.20
```
coadler, line 30:
```suggestion
          go-version: 1.20.6
```

coadler (approved):

PR #147, opened 2023-08-14
feat: add `default_apps` field to `coder_agent` resource

sreya:
- Enables a template admin to configure which apps are displayed to the user in the dashboard.

fixes the provider side of https://github.com/coder/coder/issues/8033

It takes the form 
```hcl
resource "coder_agent" "dev" {
	os = "linux"
	arch = "amd64"
	display_apps {
		vscode = false
		vscode_insiders = true
		web_terminal = false
		port_forwarding_helper = false
		ssh_helper = false
	} 
}
```

- If you do not add a `display_apps` block you opt-in to the defaults.
- To opt out of all apps you must explicitly set them all to `false`.


docs/resources/agent.md
```diff
@@ -48,6 +48,7 @@ resource "kubernetes_pod" "dev" {
 
 - `auth` (String) The authentication type the agent will use. Must be one of: "token", "google-instance-identity", "aws-instance-identity", "azure-instance-identity".
 - `connection_timeout` (Number) Time in seconds until the agent is marked as timed out when a connection with the server cannot be established. A value of zero never marks the agent as timed out.
+- `default_apps` (List of String) The list of built-in apps to display in the UI. Defaults to all apps.
```
matifali, line 51: Thanks for this. This would be a great addition.
Should we also document a list of all available built-in `default-apps`?
We should document this somewhere in our docs. Probably on this page https://coder.com/docs/v2/latest/templates

kylecarbs (approved):

docs/resources/agent.md
sreya: We can, the provider is self-documenting now that they're enumerated.

provider/agent_test.go
```diff
@@ -247,3 +248,106 @@ func TestAgent_Metadata(t *testing.T) {
...
+				// Test the fields with non-default values.
+				Config: `
+					provider "coder" {
+						url = "https://example.com"
+					}
+					resource "coder_agent" "dev" {
+						os = "linux"
+						arch = "amd64"
+						display_apps {
+							vscode = false
```
mtojek, line 270: Do we need an extra test case to check .tf with only a few apps selected:

```
display_apps {
   vscode = true
   web_terminal = true
```

provider/agent_test.go
```diff
@@ -247,3 +248,106 @@ func TestAgent_Metadata(t *testing.T) {
...
+	t.Parallel()
+	t.Run("OK", func(t *testing.T) {
+		resource.Test(t, resource.TestCase{
+			Providers: map[string]*schema.Provider{
+				"coder": provider.New(),
+			},
+			IsUnitTest: true,
+			Steps: []resource.TestStep{{
+				// Test the fields with non-default values.
+				Config: `
```
mtojek, line 262: You may want to add `display_apps` to `examples/` too.

provider/agent_test.go
sreya: Since they all have default values I tested that specifying their non-defaults (in `OK`) and then omitting them to produce the defaults (in `Omitted`) produces the desired result. I added another one anyway but I don't think it adds much value.

provider/agent.go
```diff
@@ -23,13 +23,44 @@ func agentResource() *schema.Resource {
 			if err != nil {
 				return diag.FromErr(err)
 			}
+
+			if _, ok := resourceData.GetOk("display_apps"); !ok {
```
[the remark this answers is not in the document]
sreya: The provider flags any unexpected field, but I added a test case to ensure that that is the case.

provider/examples_test.go
```diff
@@ -4,15 +4,28 @@ import (
...
 	t.Parallel()
 
+	t.Run("coder_agent", func(t *testing.T) {
+		resource.Test(t, resource.TestCase{
+			Providers: map[string]*schema.Provider{
+				"coder": provider.New(),
+			},
+			IsUnitTest: true,
+			Steps: []resource.TestStep{{
+				Config: mustReadFile(t, "../examples/resources/coder_parameter/resource.tf"),
```
mtojek, line 24: coder_agent?

mtojek (approved):
👍  

just take a look at the comment in `examples_test.go`

provider/examples_test.go
sreya: 🤦

PR #154, opened 2023-09-07
feat: add `coder_script`

kylecarbs:
See coder/coder#9287

I decided to drop the `agent` portion from the name. It seemed inconsistent with `coder_app`.

Wait for the PR into `coder/coder` before thoroughly reviewing.


docs/resources/script.md
```diff
@@ -0,0 +1,33 @@
...
+## Schema
+
+### Required
+
+- `agent_id` (String) The "id" property of a "coder_agent" resource to associate with.
+- `script` (String) The script to run.
+
+### Optional
+
+- `login_before_ready` (Boolean) This option defines whether or not the user can (by default) login to the workspace before it is ready. Ready means that e.g. the script is done and has exited. When enabled, users may see an incomplete workspace when logging in.
```
bpmct, line 25: 🤔 Why do we have this? I guess we're replacing `startup_script_behavior`? 

Could we do `behavior=blocking`?

docs/resources/script.md
```diff
@@ -0,0 +1,33 @@
...
+- `run_on_start` (Boolean) This option defines whether or not the script should run when the agent starts.
+- `run_on_stop` (Boolean) This option defines whether or not the script should run when the agent stops.
+- `schedule` (String) The schedule to run the script on. This is a cron expression.
+- `timeout` (Number) Time in seconds until the agent lifecycle status is marked as timed out, this happens when the script has not completed (exited) in the given time.
```
bpmct, line 29: Can you document what you set this to if it's a long-running process? Or I guess we want people to ampersand things `&`

bpmct (approved):

provider/script.go
```diff
@@ -0,0 +1,91 @@
...
+			},
+			"icon": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Optional: true,
+				Description: "A URL to an icon that will display in the dashboard. View built-in " +
+					"icons here: https://github.com/coder/coder/tree/main/site/static/icon. Use a " +
+					"built-in icon with `data.coder_workspace.me.access_url + \"/icon/<path>\"`.",
+			},
+			"source": {
```
sreya, line 48: The docs say `scripts` but this says `source`. I prefer `script` tbh, especially if it's supposed to be an inline script as opposed to an existing file.

sreya (approved):

docs/resources/script.md
kylecarbs: It is replacing that. The `behavior=blocking` is kinda weird, because there's also `cron`.

I could do `login_behavior=after_start`, but I'd rather just have it as a bool with the key.

docs/resources/script.md
kylecarbs: I'm going to allow long-running operations that consistently log, so I'll update this.

provider/script.go
kylecarbs: Hmm, fair. Source could mean like GitHub or something wack, I'll change it back

docs/resources/script.md
```diff
@@ -0,0 +1,36 @@
...
+### Required
+
+- `agent_id` (String) The "id" property of a "coder_agent" resource to associate with.
+- `display_name` (String) The display name of the script to display logs in the dashboard.
+- `script` (String) The script to run.
+
+### Optional
+
+- `cron` (String) The cron schedule to run the script on. This is a cron expression.
+- `icon` (String) A URL to an icon that will display in the dashboard. View built-in icons here: https://github.com/coder/coder/tree/main/site/static/icon. Use a built-in icon with `data.coder_workspace.me.access_url + "/icon/<path>"`.
```
matifali, line 27: Why do we have icon? Do we want to display all of them on the dashboard?

matifali (commented):
A few comments about displaying `coder_script` resources in the dashboard.

docs/resources/script.md
kylecarbs: Yes. These scripts will be displayed in the dashboard.

provider/script.go
```diff
@@ -0,0 +1,91 @@
...
+				Description: "This option defines whether or not the script should run when the agent starts.",
+			},
+			"run_on_stop": {
+				Type:        schema.TypeBool,
+				Default:     false,
+				ForceNew:    true,
+				Optional:    true,
+				Description: "This option defines whether or not the script should run when the agent stops.",
+			},
+			"timeout": {
```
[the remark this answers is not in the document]
kylecarbs: Yes. You'd want to add a very lengthy timeout in that case, or not have one at all.

Right now, when two CRONs overlap, the prior process won't end.

matifali:
Closes https://github.com/coder/coder/issues/9615

PR #164, opened 2023-10-10
feat: add `coder_file` resource

kylecarbs:
This will allow modules to write executable files inside of a workspace.


ammario:
We're abandoning this, right?

kylecarbs:
Yeah, this should just be done with a script.

PR #172, opened 2023-11-21
fix: add error detail for duplicate agent metadata key

Kira-Pilot:
resolves [#9713](https://github.com/coder/coder/issues/9713)

### DIFF
--- a/provider/agent.go
+++ b/provider/agent.go
@@ -47,7 +47,7 @@ func agentResource() *schema.Resource {
 				key := valueAsString(item.GetAttr("key"))
 				_, exists := itemKeys[key]
 				if exists {
-					return errorAsDiagnostics(xerrors.Errorf("duplicate agent metadata key %q", key))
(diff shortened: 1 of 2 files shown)


provider/agent.go
```diff
@@ -38,6 +39,19 @@ func agentResource() *schema.Resource {
...
 			}
+
+			rawPlan := resourceData.GetRawPlan()
+			items := rawPlan.GetAttr("metadata").AsValueSlice()
+			itemKeys := map[string]struct{}{}
+			for _, item := range items {
+				key := valueAsString(item.GetAttr("key"))
+				_, exists := itemKeys[key]
+				if exists {
+					return errorAsDiagnostics(xerrors.Errorf("duplicate agent metadata key %q", key))
```
mafredri, line 50: Not sure why we have this function, it's basically a copy of a helper from the lib. So I would suggest this change:

```suggestion
					return diag.FromErr(xerrors.Errorf("duplicate agent metadata key %q", key))
```

PR #174, opened 2023-12-08
feat: add `coder_env`

### LINKED ISSUE #170: Feature Request: Add `coder_env` Resource for Injecting Environment Variables into Workspaces
## Resource Name
`coder_env`

### Description
This feature would introduce a new resource, `coder_env`, to the Coder Terraform provider, allowing users to inject environment variables dynamically into workspaces at creation and update.

## Expected Behavior
- Enable defining environment variables in the Terraform configuration file using `coder_env`.
- Inject environment variables into the workspace specified by the `agent_id`.

## Use Case
Automating workspace setups to include necessary environment variables is essential for preparing development environments with all the configurations needed for applications and services to function correctly. An example use case is configuring go to use Artifactory, which requires setting `GOPROXY`.

### Single variable
```hcl
resource "coder_env" "example" {
  agent_id = coder_agent.example.id
  name     = "VAR1"
  value    = "Value1"
}
```
### Multiple variables
```hcl
resource "coder_env" "example" {
  agent_id = coder_agent.example.id
  item {
    name     = "VAR1"
    value    = "Value1"
  }
  item {
    name     = "VAR2"
    value    = "Value2"
  }
```
or
```hcl
resource "coder_env" "example" {
  agent_id = coder_agent.example.id
  env = {
	VAR1 = "value1"
	VAR2 = "value2"
  }
}
```

mafredri:
I opted for the single-var approach (vs multi as described in #170) because I don't see a use-case for setting _a lot_ of environment variables, and I feel it better matches the resource name and I foresee use-cases for per-environment settings (overwrite, unset, prepend, append, etc.).

Fixes #170
Refs https://github.com/coder/coder/issues/10166


provider/env.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func envResource() *schema.Resource {
+	return &schema.Resource{
+		Description: "Use this resource to set an environment variable in a workspace.",
+		CreateContext: func(_ context.Context, rd *schema.ResourceData, _ interface{}) diag.Diagnostics {
+			rd.SetId(uuid.NewString())
```
mafredri, line 15: I considered disallowing envs here, like `PATH`, `CODER_AGENT_TOKEN`, etc. However, I opted for doing this agent side instead (we'll log an error).

provider/env_test.go
```diff
@@ -0,0 +1,71 @@
...
+				} {
+					require.Equal(t, expected, script.Primary.Attributes[key])
+				}
+				return nil
+			},
+		}},
+	})
+}
+
+func TestEnvNoName(t *testing.T) {
```
mtojek, line 52: For completion, you need one more test, when the ENV variable is clear (no value).

provider/provider.go
```diff
@@ -80,6 +80,7 @@ func New() *schema.Provider {
 			"coder_app":            appResource(),
 			"coder_metadata":       metadataResource(),
 			"coder_script":         scriptResource(),
+			"coder_env":            envResource(),
```
mtojek, line 83: Thinking loud: Do we need an "opposite" resource? `coder_no_env` to remove the ENV variable?

provider/env.go
```diff
@@ -0,0 +1,42 @@
...
+				ForceNew:    true,
+				Required:    true,
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Description: "The name of the environment variable.",
+				ForceNew:    true,
+				Required:    true,
+			},
+			"value": {
```
mtojek, line 34: Thinking loud: `default_value` if not set?

provider/env.go
```diff
@@ -0,0 +1,42 @@
...
+		ReadContext:   schema.NoopContext,
+		DeleteContext: schema.NoopContext,
+		Schema: map[string]*schema.Schema{
+			"agent_id": {
+				Type:        schema.TypeString,
+				Description: `The "id" property of a "coder_agent" resource to associate with.`,
+				ForceNew:    true,
+				Required:    true,
+			},
+			"name": {
```
mtojek, line 28: Should we validate the ENV name format? Capitalized letters with underscores and digits?

mtojek (approved):
I left a few ideas for further improvement. What you proposed is good for MVP.

provider/provider.go
mafredri: For sure, I'd say that goes as a prop: `resource "coder_env" "" { name = "FOO", unset = true }`. Discussion ongoing in https://github.com/coder/coder/issues/10166.

provider/env.go
mafredri: Good idea! 👍🏻

provider/env_test.go
mafredri: I can add it, but what behavior specifically did you mean we should test/validate?

provider/env.go
mtojek: Something similar to: using generic dot files if the user didn't provide a custom URL. But I agree to postpone it for later until we meet scenario.

mtojek: This is fine for now, thanks for challenging

PR #187, opened 2024-02-07
feat: add `order` property to agent_metadata

mtojek:
Related: https://github.com/coder/coder/issues/8487

This PR adds `order` property to `coder_metadata`.

### DIFF
--- a/docs/resources/agent.md
+++ b/docs/resources/agent.md
@@ -97,4 +97,5 @@ Required:
 Optional:
 
 - `display_name` (String) The user-facing name of this value.
+- `order` (Number) The order determines the position of agent metadata in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by key (ascending order).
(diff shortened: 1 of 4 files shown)


johnstcn (approved):

docs/resources/agent.md
```diff
@@ -97,4 +114,5 @@ Required:
 Optional:
 
 - `display_name` (String) The user-facing name of this value.
+- `order` (Number) The order determines the position of agent metadata in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by key (ascending order).
```
matifali, line 117: Should we not just disallow parameters with same order?

mtojek: I wouldn't complicate things here. Imagine you have 100 metadata blocks, 1, 2,... 100. Now, you want to insert 57. With a rule forbidding duplicates, you would have to rename the rest. In this case, you can leave the duplicate.

PR #191, opened 2024-02-09
fix: improve coder_script related docs

mafredri:
- fix: improve coder_script related docs
- ~~fix: add extra validation for start_blocks_login~~ (#192)
- make gen


provider/script.go
```diff
@@ -16,15 +16,19 @@ var ScriptCRONParser = cron.NewParser(cron.Second | cron.Minute | cron.Hour | cr
...
+			startBlocksLogin, _ := rd.Get("start_blocks_login").(bool)
 			runOnStop, _ := rd.Get("run_on_stop").(bool)
 			cron, _ := rd.Get("cron").(string)
 
 			if !runOnStart && !runOnStop && cron == "" {
 				return diag.Errorf("at least one of run_on_start, run_on_stop, or cron must be set")
 			}
+			if !runOnStart && startBlocksLogin {
+				return diag.Errorf("start_blocks_login can only be set if run_on_start is true")
+			}
```
johnstcn, line 31: Does this extra check risk breaking any existing usage?

mafredri: Possibly, but it also alerts users to a misconfiguration that will cause unexpected results.

johnstcn: Might be worth calling this out as a potentially breaking change then. LGTM otherwise 👍

johnstcn (approved):

provider/script.go
mafredri: I broke it out to a separate PR, this was a good callout tbh: https://github.com/coder/terraform-provider-coder/pull/192

PR #192, opened 2024-02-09
fix!: add extra validation for start_blocks_login

mafredri:
This was originally part of #191 but split out because it is a bit of a breaking change, but hopefully it will help users avoid confusion when one is set without the other and the feature doesn't work.

This guarantee also makes our coder/coder more correct since it's common pratice to _only_ look at `start_blocks_login`.


johnstcn (approved):
I'm in favour of this change, as it will surface a misconfiguration to the user as opposed to it just silently not working.

PR #195, opened 2024-02-16
feat: relax error message for number-typed coder parameters

mtojek:
Related: https://github.com/coder/coder/issues/12192

Right now, only string-typed parameters can use custom messages. This PR relaxes strict validation rules to support custom errors for number-typed coder parameters.

### DIFF
--- a/provider/parameter.go
+++ b/provider/parameter.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/hashicorp/go-cty/cty"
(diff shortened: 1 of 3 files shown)


provider/parameter.go
```diff
@@ -466,3 +465,10 @@ func ParameterEnvironmentVariable(name string) string {
 	sum := sha256.Sum256([]byte(name))
 	return "CODER_PARAMETER_" + hex.EncodeToString(sum[:])
 }
+
+func validationError(customMessage, defaultMessage string, a ...any) error {
+	if customMessage != "" {
+		return fmt.Errorf(customMessage)
```
johnstcn, line 471: Do you anticipate template authors to pass format specifiers into this message?

provider/parameter.go
```diff
@@ -438,16 +437,16 @@ func (v *Validation) Valid(typ, value string) error {
...
-			return fmt.Errorf("value %d is less than the minimum %d", num, v.Min)
+			return validationError(v.Error, "value %d is less than the minimum %d", num, v.Min)
 		}
 		if !v.MaxDisabled && num > v.Max {
-			return fmt.Errorf("value %d is more than the maximum %d", num, v.Max)
+			return validationError(v.Error, "value %d is more than the maximum %d", num, v.Max)
 		}
 		if v.Monotonic != "" && v.Monotonic != ValidationMonotonicIncreasing && v.Monotonic != ValidationMonotonicDecreasing {
-			return fmt.Errorf("number monotonicity can be either %q or %q", ValidationMonotonicIncreasing, ValidationMonotonicDecreasing)
+			return validationError(v.Error, "number monotonicity can be either %q or %q", ValidationMonotonicIncreasing, ValidationMonotonicDecreasing)
```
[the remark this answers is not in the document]
mtojek: Oh, maybe you're right. I will disable it in this specific case.

provider/parameter.go
mtojek: I thought about support for `fmt.Sprintf`, but rather in coderd. Something like: `{min}` and `{max}`. WDYT?

johnstcn: I'd add `{value}` as well to `{min}` and `{max}`, but that makes sense to me 👍

provider/parameter.go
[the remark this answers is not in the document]
mtojek: Done

johnstcn (approved):

provider/parameter.go
mtojek: `{value}` would be useful too! I'm wondering whether the message should be rendered in the provider code, or in `coderd` 🤔

johnstcn: To be honest, provider seems the logical place as this is is closest to where it's used.

johnstcn (approved):
LGTM 👍  Up to you if you want to mention `{value}`, `{min}`, and `{max}` in the doc.

PR #202, opened 2024-03-22
feat: allow `validation` to be used with `option`s

dannykopping:
Addresses https://github.com/coder/coder/issues/11579, PR in coder/coder will close

Currently we have a conflict declared between `option` and `validation` configs in `coder_parameter` Terraform resources (@mtojek do you have some historical context as to why that was the case?).

This PR removes that conflict since there is nothing inherent in the codebase disallowing them, and we need to allow both to to accommodate the use-case mentioned in the linked issue.

I expanded the tests and added documentation for testing the provider locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,3 +36,4 @@ to setup your local Terraform to use your local version rather than the registry
    2. Run `terraform init` and observe a warning like `Warning: Provider development overrides are in effect`
 4. Run `go build -o terraform-provider-coder` to build the provider binary, which Terraform will try locate and execute
 5. All local Terraform runs will now use your local provider!
+6. _**NOTE**: we vendor in this provider into `github.com/coder/coder`, so if you're testing with a local clone then you should also run `go mod edit -replace github.com/coder/terraform-provider-coder=/path/to/terraform-provider-coder` in your clone._
--- a/README.md
+++ b/README.md
@@ -1,3 +1,38 @@
 # terraform-provider-coder
 
-See [Coder](https://github.com/coder/coder).
+Terraform provider for [Coder](https://github.com/coder/coder).
+
+### Developing
+
+Follow the instructions outlined in the [Terraform documentation](https://developer.hashicorp.com/terraform/cli/config/config-file#development-overrides-for-provider-developers)
+to setup your local Terraform to use your local version rather than the registry version.
+
+1. Create a file named `.terraformrc` in your `$HOME` directory
+2. Add the following content:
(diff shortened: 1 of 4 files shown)


mtojek:
> Currently we have a conflict declared between option and validation configs in coder_parameter Terraform resources (@mtojek do you have some historical context as to why that was the case?).

Yes, the initial implementation just assumed so. I'm sure the constraint can be relaxed now.

README.md
```diff
@@ -1,3 +1,39 @@
...
+      ```hcl
+      terraform {
+          required_providers {
+              coder = {
+                  source = "coder/coder"
+              }
+          }
+      }
+      ```
+   2. Run `terraform init` and observe a warning like `Warning: Provider development overrides are in effect`
```
mtojek, line 36: Nice trick!

provider/parameter_test.go
```diff
@@ -424,6 +440,54 @@ data "coder_parameter" "region" {
 				require.Equal(t, expected, state.Primary.Attributes[key])
 			}
 		},
+	}, {
+		Name: "NumberValidation_MonotonicWithOptions",
+		Config: `
+			data "coder_parameter" "region" {
+			  name        = "Region"
+              type        = "number"
```
mtojek, line 448: nit: formatting?

provider/parameter_test.go
```diff
@@ -424,6 +440,54 @@ data "coder_parameter" "region" {
...
+				value = 2
+			  }
+
+			  option {
+				name  = "Large"
+				value = 3
+			  }
+
+			  validation {
+			    monotonic = "increasing"
```
mtojek, line 470: nit: indent

provider/parameter_test.go
```diff
@@ -424,6 +440,54 @@ data "coder_parameter" "region" {
 				require.Equal(t, expected, state.Primary.Attributes[key])
 			}
 		},
+	}, {
+		Name: "NumberValidation_MonotonicWithOptions",
```
mtojek, line 444: I used to drop an example [here](https://github.com/coder/terraform-provider-coder/blob/main/examples/resources/coder_parameter/resource.tf) as well.

provider/parameter_test.go
dannykopping: Oh weird, my editor is probably handling tabs in a strange way. Thanks 👍

mtojek (approved):
Great!

PR #209, opened 2024-04-11
fix: set owner name and email to "default"

matifali:
Closes coder/modules#162

When the terraform plan is run, the `data.coder_workspace.me.owner_email` and `data.coder_workspace.me.owner_name` fields are blank, which causes the template build to fail. 

This should resolve the issue by setting those entries to a non-empty string.

### DIFF
--- a/provider/workspace.go
+++ b/provider/workspace.go
@@ -35,6 +35,9 @@ func workspaceDataSource() *schema.Resource {
 			_ = rd.Set("owner", owner)
 
 			ownerEmail := os.Getenv("CODER_WORKSPACE_OWNER_EMAIL")
+			if ownerEmail == "" {
+				ownerEmail = "default"
+			}
 			_ = rd.Set("owner_email", ownerEmail)
 
 			ownerGroupsText := os.Getenv("CODER_WORKSPACE_OWNER_GROUPS")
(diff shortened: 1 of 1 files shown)


provider/workspace.go
```diff
@@ -35,6 +35,9 @@ func workspaceDataSource() *schema.Resource {
 			_ = rd.Set("owner", owner)
 
 			ownerEmail := os.Getenv("CODER_WORKSPACE_OWNER_EMAIL")
+			if ownerEmail == "" {
+				ownerEmail = "default"
```
mtojek, line 39: would it be safer to set it to `default@domain.tld` or something similar?

matifali: Yes I think so. We can update.

johnstcn: `example.com` is also a good placeholder for a domain name.

johnstcn (commented):
.We should add a unit test to validate this behaviour, and we should probably also validate that this only happens during the plan phase; if these values are unset at the apply stage then this is a pretty big bug somewhere in the coder <-> provider interface.

matifali:
> We should add a unit test to validate this behaviour, and we should probably also validate that this only happens during the plan phase; if these values are unset at the apply stage, then this is a pretty big bug somewhere in the coder <-> provider interface.

I would be happy if someone could take this from here and add the tests. I opened this, considering its a smaller change. I am yet to feel confident about my Go skills.

johnstcn (approved):

PR #223, opened 2024-05-10
feat: support workspace tags

mtojek:
Related: https://github.com/coder/coder/issues/13218

This PR adds Terraform definition for `coder_workspace_tags` with resource reference tests.

### DIFF
--- a/provider/workspace_tags.go
+++ b/provider/workspace_tags.go
@@ -8,6 +8,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+const TerraformWorkDirEnv = "CODER_TF_WORK_DIR"
+
 type WorkspaceTags struct {
 	Tag []Tag
 }
--- a/provider/workspace_tags.go
+++ b/provider/workspace_tags.go
@@ -2,7 +2,6 @@ package provider
 
 import (
 	"context"
-	"log"
 
 	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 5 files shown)


provider/workspace_tags.go
```diff
@@ -0,0 +1,53 @@
...
+type Tag struct {
+	Name  string
+	Value string
+}
+
+func workspaceTagDataSource() *schema.Resource {
+	return &schema.Resource{
+		Description: "Use this data source to configure workspace tags to select provisioners.",
+		ReadContext: func(ctx context.Context, rd *schema.ResourceData, i interface{}) diag.Diagnostics {
+			rd.SetId(uuid.NewString())
```
mtojek, line 24: For reviewers:

Originally I wanted to place here the logic to load HCL files and read raw content of `value` properties. Unfortunately, I hit an interface blocker. Here is the story:
1. Pass `CODER_TF_WORK_DIR` from provisioner with disk path to the template.
2. Use [terraform-config-inspect](https://github.com/hashicorp/terraform-config-inspect) to load module and get references (filename + pos) to data resources.
3. Read raw attributes of the relevant data resource.

BUT...

`schema.ResourceData` is not aware of its own identity. It knows only its child properties, raw config and state diff. I'm afraid this is a blocker we can't find a workaround for, and we have to implement this logic in coderd.

spikecurtis: I don't understand why we want or need to do this here.  This gets called at _build_ time to allow other terraform resources to read data.  I think we can just return the actual tag _values_ and call it a day.

The loading of the files and parsing HCL should happen during the template version import, where we already have access to the raw files: https://github.com/coder/coder/blob/f14927955d7361d9a50c633e69fbebca2d9946cb/provisioner/terraform/parse.go#L26

johnstcn (approved):

Issue #222, opened 2024-05-09
bump go version


matifali:
Extracted from https://github.com/coder/terraform-provider-coder/issues/217#issuecomment-2102217213

As per https://endoflife.date/go, only the latest 2 versions of Go are supported for any security patches. `terraform-terraform-coder` is on 1.20, which has been unsupported for the last 3 months. So, probably should consider bumping the go version to 1.22.

johnstcn:
cc @mtojek -- candidate for between-sprints?

mtojek:
definitely

PR #228, opened 2024-05-23
chore: add devcontainer config

johnstcn:
For convenience!

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,6 @@
   "image": "mcr.microsoft.com/devcontainers/go:1.20",
   "features": {
     "ghcr.io/devcontainers/features/terraform:1": {
-      "version": "1.7.5",
       "installTerraformDocs": true
     }
   }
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "envbuilder",
```
mafredri, line 2: Was this supposed to be something else?

johnstcn: copy-pasta 😅

mafredri (approved):

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "terraform-provider-coder",
+  "image": "mcr.microsoft.com/devcontainers/go:1.20",
```
mafredri, line 3: Do we still use Go 1.20 here, btw? Should we match coder/coder or go 1.22? 🤔

johnstcn: https://github.com/coder/terraform-provider-coder/issues/222

PR #230, opened 2024-05-23
feat: add coder_workspace_owner datasource

johnstcn:
NOTE: still needs some coder/coder changes.

Adds a `coder_workspace_owner` data source populated from all available information, if present.

Deprecates relevant fields of `coder_workspace`.


provider/provider.go
```diff
@@ -74,6 +74,7 @@ func New() *schema.Provider {
 			"coder_parameter":      parameterDataSource(),
 			"coder_git_auth":       gitAuthDataSource(),
 			"coder_external_auth":  externalAuthDataSource(),
+			"coder_user":           userDataSource(),
```
johnstcn, line 77: This will only ever be populated with information regarding the workspace owner.
Should I rename it to `coder_workspace_owner` instead?

provider/provider.go
```diff
@@ -74,6 +74,7 @@ func New() *schema.Provider {
 			"coder_parameter":      parameterDataSource(),
 			"coder_git_auth":       gitAuthDataSource(),
 			"coder_external_auth":  externalAuthDataSource(),
+			"coder_user":           workspaceOwnerDataSource(),
```
johnstcn, line 77: should be coder_workspace_owner, will fix

matifali:
@stirby if we are planning to remove all deprecated items in 1.0 then I suggest we should list them in a milestone or an ☔ issue.

PR #232, opened 2024-05-24
fix: provider: coder_workspace_owner: avoid setting null values

johnstcn:
Follow-up of https://github.com/coder/terraform-provider-coder/pull/230


mafredri (approved):
LGTM!

PR #233, opened 2024-05-28
chore: add integration test for provider

johnstcn:
This PR adds integration tests for the provider against an arbitrary version of Coder.

Example usage:
```
CODER_VERSION="v2.10.0-devel-3a9a7d199-amd64" go test -v ./integration
```

Testing methodology:
- Create an ephemeral Coder instance running in a container
- Perform initial setup
- Import a predefined template that creates a single file
- Create a workspace that populates the file with the actual output
- Compare the actual versus expected output


go.mod
```diff
@@ -1,8 +1,11 @@
 module github.com/coder/terraform-provider-coder
 
-go 1.20
+go 1.21
+
+toolchain go1.22.3
```
johnstcn, line 5: This was not a deliberate change and is likely due to importing `docker/docker`.

dannykopping (commented):
Looking super slick, added a few notes

integration/integration_test.go
```diff
@@ -141,24 +179,39 @@ func setup(ctx context.Context, t *testing.T) string {
...
+	// nolint:gosec // For testing only.
 	var (
 		testEmail    = "testing@coder.com"
 		testPassword = "InsecurePassw0rd!"
 		testUsername = "testing"
 	)
 
 	// Wait for container to come up
-	execContainer(ctx, t, ctr.ID, `until curl -s --fail http://localhost:3000/healthz; do sleep 1; done`)
+waitLoop:
```
dannykopping, line 190: I think `require.Eventually` might be more idiomatic?

integration/test-data-source/main.tf
```diff
@@ -0,0 +1,65 @@
...
+    coder = {
+      source = "coder/coder"
+    }
+    local = {
+      source = "hashicorp/local"
+    }
+  }
+}
+
+// TODO: test coder_external_auth and coder_git_auth
```
matifali, line 12: `coder_git_auth` could be skipped as it has been deprecated for a while.

integration/test-data-source/main.tf
```diff
@@ -0,0 +1,65 @@
+terraform {
```
mtojek, line 1: nit: curious if we can use a file from the examples directory, or put it there.

mtojek (approved):
Nice testing improvement, so let's get it in.

integration/test-data-source/main.tf
johnstcn: We could use some, but most have dependencies like gcloud credentials that we can't necessarily provide.

johnstcn:
Will add to GitHub testing workflow in a separate PR.

Issue #242, opened 2024-06-24
Code example not consumed in docs generation


stirby:
All of our resources and data sources should have code examples. While some are missing entirely, [`coder_parameter`](https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/parameter) has an [example](https://github.com/coder/terraform-provider-coder/blob/main/examples/resources/coder_parameter/resource.tf) which does not appear in the documentation.

Possibly an issue with the [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) generation call.

coder_parameter is the only resource or data source with an example that doesn't propagate to the docs.

stirby:
Pathing issue, not a problem with Docs generation. I'll fix this.

stirby:
coder_workspace_tags also subject. 

Interestingly, they are the only examples prefixed with:

```hcl
provider "coder" {}

...
```

PR #243, opened 2024-06-24
chore(docs): move coder_parameter example

stirby:
The `coder_parameter` **data-source**'s code example was in the **resources** directory by mistake, moved.

Fixes https://github.com/coder/terraform-provider-coder/issues/242.

---

Curious what the failed `examples_test.go` looks like so intentionally ignored.


stirby:
@mtojek I removed the `coder_parameter` test from `examples_test.go` since that test targets resources specifically. 

I find it interesting that the only two resources which don't generate code examples in the docs are prefixed with:

```hcl
provider "coder" {}
```

Since `coder_workspace_tags` is in the correct directory, the problem may be deeper than a pathing issue.

I couldn't find any errors in the output of `make gen`'s `tfplugindocs` call:

```
...
generating missing resource content
generating new template for "coder_app"
...
generating missing data source content
...
generating new template for data-source "coder_workspace_owner"
generating new template for data-source "coder_workspace_tags"
...
generating new template for data-source "coder_parameter"
generating missing function content
generating missing provider content
...
```

I also tried to add other resources to the `examples_test.go`, each of which failed. 

---

All this said, I would actually like to take a stab at rewriting the `examples_test.go`. My fear is that the tests only pass with inclusion of the provider, which prevents the docs from correctly generating.

Mind taking a look?

stirby:
For now, lazily duplicated resource examples test loop for data sources to test.

mtojek:
Alright, I think I made the CI happy now. `coder_parameter` and `coder_workspace_tags` are in `data-sources` now, and `examples_test.go` is validating them.

> I also tried to add other resources to the examples_test.go, each of which failed.

This may not be straightforward if examples depend on external fields or providers. For instance: `coder_agent_instance` uses `google_compute_instance` and we do not import `gcp` module.

mtojek (approved):
👍

Issue #237, opened 2024-06-03
docs missing deprecation notice and migration steps


matifali:
Provider docs at registry.coder.com do not mention the deprecation of parameters. 

For example, multiple fields of `data.coder_workspace` were deprecated in `v0.23.0`, but the docs at https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/workspace do not indicate any such information.

Additionally, when a resource is deprecated, the message should contain the new resource to be migrated to or a link to some migration guide both in Terraform logs and on the docs page.

matifali:
related #203

johnstcn:
Looks like you need to specify `DeprecationMessage` and not just `Deprecated`.
https://developer.hashicorp.com/terraform/plugin/framework/handling-data/schemas#deprecationmessage-1

matifali:
Good find @johnstcn 🧠

mtojek:
Ok, I researched the issue. It may not be a straightforward solution or, in other words, it could be easier to apply patch `.md` files.

The [docs generator](https://github.com/hashicorp/terraform-plugin-docs) strongly depends on `providers schema`:

<img width="1086" alt="Screenshot 2024-06-25 at 16 02 38" src="https://github.com/coder/terraform-provider-coder/assets/14044910/d6edd0e0-664f-4dd1-aaec-e846f615d702">... and the schema does not expose the deprecation message.

It would be hard/impossible to expose the message via the fork&modify approach, so my recommendation is to implement a script which will run after the hashicorp generator, and "patch" docs:

```
- `owner_email` (String, Deprecated) Email address of the workspace owner.
```

->

```
- `owner_email` (String, Deprecated: Use `coder_workspace_owner.email` instead.) Email address of the workspace owner.
```

@johnstcn @stirby Please let me know your thoughts 👍

johnstcn:
Thanks for investigating! I think that's a perfectly valid workaround.

PR #245, opened 2024-06-26
feat: include deprecation message in docs

mtojek:
Fixes: https://github.com/coder/terraform-provider-coder/issues/237

This PR adds a script to patch Terraform docs to contain the deprecation message.


scripts/docsgen/main.go
```diff
@@ -0,0 +1,88 @@
...
+	"strings"
+
+	"github.com/coder/terraform-provider-coder/provider"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// This script patches Markdown docs generated by `terraform-plugin-docs` to expose the original deprecation message.
+
+var (
+	reDeprecatedProperty = regexp.MustCompile("`([^`]+)` \\(([^,\\)]+), Deprecated\\) ([^\n]+)")
```
johnstcn, line 19: readability: use backticks for regexes so you don't need to escape all the things

scripts/docsgen/main.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func writeDeprecationMessage(doc []byte, schemas map[string]*schema.Schema) []byte {
+	return reDeprecatedProperty.ReplaceAllFunc(doc, func(m []byte) []byte {
+		matches := reDeprecatedProperty.FindSubmatch(m)
+		propertyName := matches[1]
+		description := matches[3]
+
+		sch := schemas[string(propertyName)]
+		if string(description) != sch.Description {
```
johnstcn, line 83: nit: probably no harm to drop a log here for troubleshooting

scripts/docsgen/main.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func exposeDeprecationMessage(p *schema.Provider) error {
+	// Patch data-sources
+	for dataSourceName, dataSource := range p.DataSourcesMap {
+		docFile := filepath.Join("docs", "data-sources", strings.Replace(dataSourceName, "coder_", "", 1)+".md")
+
+		err := adjustDocFile(docFile, dataSource.Schema)
+		if err != nil {
+			return err
```
johnstcn, line 38: wrap error

johnstcn (approved):
Thanks for figuring this out! 
Just a couple of small changes I'd like to see in the docsgen script to make it easier to debug changes in future.

scripts/docsgen/main.go
mtojek: no so easy here? we have backticks in the regexp already, and we can't escape them

johnstcn: 🤦 never mind then!

scripts/docsgen/main.go
mtojek: Fixed

scripts/docsgen/main.go
mtojek: Fixed

Issue #217, opened 2024-04-29
Coder Terraform Provider v1.0 Docs 


stirby:
Customers have requested that we promote our terraform provider to `v1.0`. The product is quite stable and, of course, is used in every one of our customers' deployments. 

To show that this is no longer breaking, we should promote the release versioning to `v1.X.X`; before we do so, the [docs](https://registry.terraform.io/providers/coder/coder/latest/docs) should meet `v1.0` standards. As I mentioned, we don't get many stability requests, but there are often requests for elaboration on the provider's behavior. There are a few resources/data sources that are missing code examples, or may be enriched via context.

### Docs checklist for v1.0 
- [x] Resource docs
  - [x] [`coder_env`](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/env)
    - [x] Code examples
  - [x] [`coder_script`](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/script)
    - [x] Code examples
    - [x] Explanation on parallel execution 
    - [x] ~Startup script implications~
    - [x] ~Link to discussion on [execution ordering](https://github.com/coder/coder/issues/10352) (frequently requested)~
      - not necessary 
- [x] Data source docs
  - [x] [`coder_external_auth`](https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/external_auth)
    - [x] Code examples
    - [x] Link to coder.com external auth docs
  - [x] [`coder_parameter`](https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/parameter)
    - [x] Code examples (show validation types, monotonicity, etc)
      - Should just grep the closed customer requests on `coder_parameter` ([like this](https://github.com/coder/customers/issues?q=is%3Aissue+is%3Aclosed+coder_parameter+)) 
    - [x] https://github.com/coder/terraform-provider-coder/issues/242
    - [x] Optional: Fix inline link under `icon`
    - [x] ~Optional: Link to discussion on coder_parameter extension~
  - [x] [`coder_provisioner`](https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/provisioner)
    - [x] Code example
    - [x] Optional: Fix inline links
  - [x] [`coder_workspace`](https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/workspace)
    - [x] ~Optional: Expand code examples~
      - Not enough attributes to warrant further examples  
[...]

PR #246, opened 2024-06-27
feat: require environment variables

mtojek:
Fixes: https://github.com/coder/terraform-provider-coder/issues/217
Related: https://github.com/coder/coder/pull/13680

This PR enables constraint validation for required environment variables.

### DIFF
--- a/provider/helpers/env.go
+++ b/provider/helpers/env.go
@@ -6,7 +6,13 @@ import (
 )
 
 // RequireEnv requires environment variable to be present.
+// The constraint can be verified only during execution of the workspace build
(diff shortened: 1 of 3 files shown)


johnstcn (commented):
Can you update the integration test as well?

mtojek:
Yes, but I don't have a clue: how?

The integration test operates on Coder CLI, like e2e test, and this is an internal change. We would have to trick Coder to misbehave?

johnstcn:
Yeah, the more I look at this change, the less I can see an integration test being able to assert anything useful.

johnstcn (approved):

PR #247, opened 2024-06-27
chore: improve integration test compatibility with older Coder versions

johnstcn:
Previously, testing with coder versions older than 2.7 would fail due to the `template push` command not creating the template automatically.

Also, improves logic around pulling the image if it does not exist locally.


mtojek (approved):
👍

PR #253, opened 2024-07-02
docs: improve docs by using code-styled markdown

### LINKED ISSUE #203: Document `coder_git_auth` deprecation
In #163 `coder_git_auth` was marked as deprecated. It does show a message when used but it would be helpful to provide that same information in the documentation. Right now the documentation doesn't say anything about the deprecated state of `coder_git_auth` or mentions you should use `coder_external_auth` instead.

In fact because of the fact that the `coder_external_auth` mentions multiple services but not anything like git or GitHub (gh is mentioned but unless you know it you may not make the connection) people may be lead to believe they should use `coder_git_auth` for git authentication and `coder_external_auth` for authentication with other services. This may lead people to use one and not realize the deprecation until they first use it in a template. Unless like me they go looking for what the difference is and stumble on the PR.

matifali:
Makes use of backticks to format resources and data sources names in docs.
Also resolves a mix use of backticked and quoted strings.

The format I tried to follow is:
- all resources and data sources `names` and their properties' `names` re formated as code-blocks. e.g., `coder_agent`, `startup_script` etc. 
- all `"values"` of any property are formatted as quoted code blocks. e.g., for `share` the values are `"owner"`, `"authenticated"` and `"public"`. 
- `true` and `false` are formated as code-blocks. 
- 
Also fixes #203 

| **Before** | **After** |
| ---------- | --------- |
|image|image|
|image|image|

> [!TIP]
> Docs previews generated at https://registry.terraform.io/tools/doc-preview


matifali:
> [!TIP]
> Docs previews generated at https://registry.terraform.io/tools/doc-preview

stirby (approved):
Great improvements to consistency and maturity. Nice addition.

mtojek (approved):
As CI is happy, I'm happy too 👍

Issue #250, opened 2024-07-01
Remove deprecated resources/data sources


matifali:
~Remove all deprecated items before #217~

Edit: We plan to remove it in v2.0.0 now.

PR #256, opened 2024-07-03
chore!: remove deprecated feature_use_managed_variables

matifali:
The `feature_use_managed_variables` option has been removed as Terraform variables are now exclusively utilized for template-wide variables. This commit removes the deprecated option from the provider code.

contributes to #250


johnstcn (approved):

johnstcn:
We will need to bump the major version after this change.

mtojek (approved):

PR #257, opened 2024-07-04
chore!: remove deprecated fields from `coder_app`  and `coder_agent` resources

matifali:
This pull request removes the deprecated fields from the `coder_app` and `coder_agent`.
contributes to #250


johnstcn (approved):

johnstcn:
We will need to bump the major version after this change.

mtojek (approved):

PR #264, opened 2024-08-01
feat: add owner_oidc_refresh_token to coder_workspace_owner data source

ffais:
Add OIDC refresh token to coder_workspace_owner data source, useful to renew access token inside the workspace without restarting it.

### DIFF
--- a/provider/workspace_owner.go
+++ b/provider/workspace_owner.go
@@ -52,6 +52,7 @@ func workspaceOwnerDataSource() *schema.Resource {
 
 			_ = rd.Set("session_token", os.Getenv("CODER_WORKSPACE_OWNER_SESSION_TOKEN"))
 			_ = rd.Set("oidc_access_token", os.Getenv("CODER_WORKSPACE_OWNER_OIDC_ACCESS_TOKEN"))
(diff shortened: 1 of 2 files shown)


johnstcn:
@ffais can you run `make gen` to update the plugin docs? This should help you pass CI.

johnstcn:
@ffais is there a particular reason you need this? Is it not possible to use the [`coder external-auth access-token` command](https://coder.com/docs/cli/external-auth_access-token) for your use-case? This should get automatically refreshed by Coder.

ffais:
The main reason is to use the identity of the user who logged into coder also to authenticate the same on another component, in the most transparent way possible for the user.

Propagating the refresh token within the workspace is the most transparent solution for the workspace user, but as we have already said in pull request [#14074](https://github.com/coder/coder/pull/14074), this could,in some scenarios, lead to premature revocation of the token.

As far as I understand, using external auth the user have to perform an additional auth flow to obtain an access token within the workspace. Is this correct?

johnstcn:
> As far as I understand, using external auth the user have to perform an additional auth flow to obtain an access token within the workspace. Is this correct?

It requires a once-off flow for the user to link their account.

https://coder.com/docs/admin/external-auth#require-git-authentication-in-templates

PR #276, opened 2024-09-05
feat: add 'hidden' field to 'coder_app' provider

DanielleMaywood:
Relates to https://github.com/coder/coder/pull/14570

Adds a `hidden` field to the `coder_app` provider.

### DIFF
--- a/provider/app_test.go
+++ b/provider/app_test.go
@@ -45,6 +45,7 @@ func TestApp(t *testing.T) {
 						threshold = 6
 					}
 					order = 4
+					hidden = false
 				}
 				`,
 				Check: func(state *terraform.State) error {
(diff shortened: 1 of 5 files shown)


mtojek (approved):
LGTM 👍

provider/app_test.go
```diff
@@ -246,4 +248,75 @@ func TestApp(t *testing.T) {
...
+				os = "linux"
+				arch = "amd64"
+			}
+			resource "coder_app" "test" {
+				agent_id = coder_agent.dev.id
+				slug = "test"
+				display_name = "Testing"
+				url = "https://google.com"
+				external = true
+				hidden = true
```
mtojek, line 273: @DanielleMaywood @johnstcn  if the app is `hidden`, does it make sense to define `display_name`, `url`, etc.? I don't know the answer, thinking loud here. If so:
* we could improve the validation to check conflicting values of properties
* we should check if coder/coder does not break as it could be expecting all of them to be defined

johnstcn: I'd say dropping a `diag.Warn` is the most we should do here.
For the latter, I believe Danielle already tested manually, but I suppose we could add an integration test that defines both with and without.

johnstcn (approved):

PR #277, opened 2024-09-05
fix: validate agent & resource metadata keys during plan

ethanndickson:
Closes https://github.com/coder/coder/issues/14387.
Closes https://github.com/coder/coder/issues/14569.

Previously, the configuration had to be applied in order for the provider to complain about duplicate keys - now just a plan will suffice.

### DIFF
--- a/provider/agent.go
+++ b/provider/agent.go
@@ -43,18 +43,6 @@ func agentResource() *schema.Resource {
 				}
 			}
 
-			rawPlan := resourceData.GetRawPlan()
-			items := rawPlan.GetAttr("metadata").AsValueSlice()
-			itemKeys := map[string]struct{}{}
-			for _, item := range items {
-				key := valueAsString(item.GetAttr("key"))
-				_, exists := itemKeys[key]
-				if exists {
-					return diag.FromErr(xerrors.Errorf("duplicate agent metadata key %q", key))
-				}
-				itemKeys[key] = struct{}{}
-			}
-
 			return updateInitScript(resourceData, i)
 		},
 		ReadWithoutTimeout: func(ctx context.Context, resourceData *schema.ResourceData, i interface{}) diag.Diagnostics {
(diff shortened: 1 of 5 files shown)


ethanndickson:
* **#277** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coder/277?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @ethanndickson and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

provider/agent.go
```diff
@@ -272,6 +260,23 @@ func agentResource() *schema.Resource {
 				Optional:    true,
 			},
 		},
+		CustomizeDiff: func(ctx context.Context, rd *schema.ResourceDiff, i interface{}) error {
+			if !rd.HasChange("metadata") {
+				return nil
+			}
+
+			keys := map[string]struct{}{}
```
johnstcn, line 268: Nit: If you change this to a `map[string]bool` you can make the below condition look a lot nicer, at the expense of a few bytes:

```
if keys[key] {
  return ...
}
```

provider/agent.go
```diff
@@ -272,6 +260,23 @@ func agentResource() *schema.Resource {
...
+			for _, t := range rd.Get("metadata").([]interface{}) {
```
johnstcn, line 269: Would it make sense to check that `rd.Get("metadata")` can be cast to a `[]interface{}`?
Same for below with `taint`.

Also, I think you can just use `any` as an alias for `interface{}` now.

provider/metadata.go
```diff
@@ -111,5 +112,22 @@ func metadataResource() *schema.Resource {
 				},
 			},
 		},
+		CustomizeDiff: func(ctx context.Context, rd *schema.ResourceDiff, i interface{}) error {
+			if !rd.HasChange("item") {
+				return nil
+			}
+
+			keys := map[string]struct{}{}
+			for _, t := range rd.Get("item").([]interface{}) {
```
johnstcn, line 121: check cast?

integration/integration_test.go
```diff
@@ -103,7 +103,7 @@ func TestIntegration(t *testing.T) {
 				"workspace.transition":              `start`,
 				"workspace_owner.email":             `testing@coder\.com`,
 				"workspace_owner.full_name":         `default`,
-				"workspace_owner.groups":            `\[\]`,
+				"workspace_owner.groups":            `\[\"Everyone\"\]`,
```
johnstcn, line 106: Integration test doesn't appear to like this. Should this be moved to a separate PR?

provider/agent.go
ethanndickson: I don't think it's possible for either cast to fail, but I'll add checks in-case the schema changes

integration/integration_test.go
ethanndickson: That's annoying, I thought this would appease.. I'll make a new PR.

johnstcn (approved):

matifali:
Do we want to port this to `release/1.0` and do a patch?

ethanndickson:
Yeah, potentially. What are the breaking changes on main not included in any release?

matifali:
We removed a couple of deprecated resources and attributes.

PR #278, opened 2024-09-05
fix: optional everyone group in integration test

ethanndickson:
CI is currently failing and blocking PRs as the integration test running against Coder v2.15.0 (mainline) adds the "Everyone" group to the `workspace_owner` `groups` field. However, the CI also runs the integration test against Coder v2.14.2 (stable), which doesn't include the everyone group. 

We'll accept either an empty groups list, or a list with the everyone group to enable running the test against both versions.


johnstcn (approved):
:+1:

Issue #235, opened 2024-05-28
feat: add login_type to coder_workspace_owner data source


johnstcn:
...we can expose account type, i.e., `password`, `github`, `oidc`, or `none` (could be extracted to a new issue) [https://github.com/coder/terraform-provider-coder/assets/10648092/58097426-aed9-4f80-bcf9-75e8ab7358c0](https://github.com/coder/terraform-provider-coder/assets/10648092/58097426-aed9-4f80-bcf9-75e8ab7358c0)

_Originally posted by @matifali in _ [_https://github.com/coder/terraform-provider-coder/issues/219#issuecomment-2135511772_](https://github.com/coder/terraform-provider-coder/issues/219#issuecomment-2135511772)

pmareke:
Hi, is this issue still open?

johnstcn:
> Hi, is this issue still open? I would like to give it a try.

Yes, this issue has not yet been worked on!

pmareke:
Great, done!

matifali:
Hi @pmareke thank your for your contribution. Could you run `make gen` to let the linter pass?

pmareke:
Done!

pmareke:
> Hi @pmareke thank your for your contribution. Could you run `make gen` to let the linter pass?

Done!

johnstcn:
https://github.com/coder/terraform-provider-coder/pull/287 updates the provider to add the field. A separate PR will be required in `coder/coder` to set `CODER_WORKSPACE_OWNER_LOGIN_TYPE`.

pmareke:
Thanks so much @johnstcn, I really like doing the PR!

dannykopping:
> [#287](https://github.com/coder/terraform-provider-coder/pull/287) updates the provider to add the field. A separate PR will be required in `coder/coder` to set `CODER_WORKSPACE_OWNER_LOGIN_TYPE`.

@johnstcn do you know if there was a subsequent PR raised for this?
Right now the provider just raised a warning each time: `Warning: WARNING: The CODER_WORKSPACE_OWNER_LOGIN_TYPE env variable is not set`.

johnstcn:
@dannykopping @matifali looks like it wasn't. I can add one quickly.

PR #287, opened 2024-09-10
Add login_type to coder_workspace_owner data source #235

pmareke:
This PR add the login type the the `coder_workspace_owner` data source (Issue #235).

I took some decisions (like the default value to `none`), but if you think this should be different please let me know.

Thanks!

### DIFF
--- a/provider/workspace_owner.go
+++ b/provider/workspace_owner.go
@@ -54,11 +54,15 @@ func workspaceOwnerDataSource() *schema.Resource {
 			_ = rd.Set("oidc_access_token", os.Getenv("CODER_WORKSPACE_OWNER_OIDC_ACCESS_TOKEN"))
 
 			if os.Getenv("CODER_WORKSPACE_OWNER_LOGIN_TYPE") == "" {
-				diag.Warn("The CODER_WORKSPACE_OWNER_LOGIN_TYPE env variable is not set")
+				diags := req.Config.Get(ctx, &rd)
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Warning,
+					Summmary: "WARNING: The CODER_WORKSPACE_OWNER_LOGIN_TYPE env variable is not set"
+				},
 			}
 			_ = rd.Set("login_type", os.Getenv("CODER_WORKSPACE_OWNER_LOGIN_TYPE"))
 
-			return nil
+			return diags
 		},
 		Schema: map[string]*schema.Schema{
 			"id": {
--- a/provider/workspace_owner.go
+++ b/provider/workspace_owner.go
@@ -3,7 +3,9 @@ package provider
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/google/uuid"
@@ -52,6 +54,17 @@ func workspaceOwnerDataSource() *schema.Resource {
 
 			_ = rd.Set("session_token", os.Getenv("CODER_WORKSPACE_OWNER_SESSION_TOKEN"))
 			_ = rd.Set("oidc_access_token", os.Getenv("CODER_WORKSPACE_OWNER_OIDC_ACCESS_TOKEN"))
+
+			types := []string{"password", "github", "oidc", "none"}
+			if login_type := os.Getenv("CODER_WORKSPACE_OWNER_LOGIN_TYPE"); login_type != "" {
+				if !slices.Contains(types, login_type) {
+					errorMessage := "invalid login type: %s, the valid types are: 'password, github, oidc, or none'"
+					return diag.Errorf(errorMessage, errors.New(errorMessage))
+				}
+				_ = rd.Set("login_type", login_type)
+			} else {
+				_ = rd.Set("login_type", "none")
+			}
 
 			return nil
 		},
@@ -106,6 +119,11 @@ func workspaceOwnerDataSource() *schema.Resource {
 				Description: "A valid OpenID Connect access token of the workspace owner. " +
 					"This is only available if the workspace owner authenticated with OpenID Connect. " +
 					"If a valid token cannot be obtained, this value will be an empty string.",
+			},
+			"login_type": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The type of login the user has.",
 			},
 		},
 	}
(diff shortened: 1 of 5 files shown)


pmareke:
Done!

provider/workspace_owner.go
```diff
@@ -53,6 +53,12 @@ func workspaceOwnerDataSource() *schema.Resource {
 			_ = rd.Set("session_token", os.Getenv("CODER_WORKSPACE_OWNER_SESSION_TOKEN"))
 			_ = rd.Set("oidc_access_token", os.Getenv("CODER_WORKSPACE_OWNER_OIDC_ACCESS_TOKEN"))
 
+			if login_type := os.Getenv("CODER_WORKSPACE_OWNER_LOGIN_TYPE"); login_type != "" {
+				_ = rd.Set("login_type", login_type)
+			} else {
+				_ = rd.Set("login_type", "none")
```
johnstcn, line 59: @mtojek should we instead set `login_type` to an empty string if Coder does not set the required env?

johnstcn: Perhaps at the very least, drop a `diag.Warn`?

mtojek: > should we instead set login_type to an empty string

This seems reasonable and consistent with other ENV vars. I would start with empty string, we can always revisit it in the future and replace it with `none`.

mtojek (commented):
Thank you for your contribution!

provider/workspace_owner.go
pmareke: > This seems reasonable and consistent with other ENV vars. I would start with empty string, we can always revisit it in the future and replace it with `none`.

Great, what about the warn?

pmareke: Sorry, but taking a look the the `diag` package there is not `Warn` method.

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/diag

johnstcn: Sorry! I should have been more specific -- you can append a `diag.Diagnostic` with severity `diag.Warning` to the second return argument of type `diag.Diagnostics`.

pmareke: Done, this is the first time I have to deal with the `diags`, so please let me know if this solution is correct 🙏🏼 .

provider/workspace_owner.go
```diff
@@ -53,7 +53,16 @@ func workspaceOwnerDataSource() *schema.Resource {
 			_ = rd.Set("session_token", os.Getenv("CODER_WORKSPACE_OWNER_SESSION_TOKEN"))
 			_ = rd.Set("oidc_access_token", os.Getenv("CODER_WORKSPACE_OWNER_OIDC_ACCESS_TOKEN"))
 
-			return nil
+			if os.Getenv("CODER_WORKSPACE_OWNER_LOGIN_TYPE") == "" {
+				diags := req.Config.Get(ctx, &rd)
+				diags = append(diags, diag.Diagnostic{
```
johnstcn, line 58: The `diags` variable you define above is not in the scope of the `return` statement on line 65, so this will not compile.

Instead, you can define a new `diags diag.Diagnostics` variable at the start of the function and reference it here.

provider/workspace_owner.go
```diff
@@ -53,7 +53,16 @@ func workspaceOwnerDataSource() *schema.Resource {
 			_ = rd.Set("session_token", os.Getenv("CODER_WORKSPACE_OWNER_SESSION_TOKEN"))
 			_ = rd.Set("oidc_access_token", os.Getenv("CODER_WORKSPACE_OWNER_OIDC_ACCESS_TOKEN"))
 
-			return nil
+			if os.Getenv("CODER_WORKSPACE_OWNER_LOGIN_TYPE") == "" {
+				diags := req.Config.Get(ctx, &rd)
```
johnstcn, line 57: There is no `req` variable in scope here, but I don't think this line is necessary (see comment on next line).

provider/workspace_owner.go
```diff
@@ -53,7 +53,16 @@ func workspaceOwnerDataSource() *schema.Resource {
...
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Warning,
+					Summmary: "WARNING: The CODER_WORKSPACE_OWNER_LOGIN_TYPE env variable is not set"
```
johnstcn, line 60: Ensure that all struct field declarations are terminated by a comma. https://go.dev/ref/spec#Struct_types

provider/workspace_owner.go
```diff
@@ -53,7 +53,16 @@ func workspaceOwnerDataSource() *schema.Resource {
...
+				},
```
johnstcn, line 61: Missing a closing bracket here for `append`. Also, trailing comma.

provider/workspace_owner.go
pmareke: Thanks and done!

I was not sure if creating a new `diag.Diagnostics` will remove the previous warnings.

pmareke:
All fixed @johnstcn thanks for your time!

provider/workspace_owner.go
```diff
@@ -53,7 +55,15 @@ func workspaceOwnerDataSource() *schema.Resource {
 			_ = rd.Set("session_token", os.Getenv("CODER_WORKSPACE_OWNER_SESSION_TOKEN"))
 			_ = rd.Set("oidc_access_token", os.Getenv("CODER_WORKSPACE_OWNER_OIDC_ACCESS_TOKEN"))
 
-			return nil
+			if os.Getenv("CODER_WORKSPACE_OWNER_LOGIN_TYPE") == "" {
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Warning,
+					Summmary: "WARNING: The CODER_WORKSPACE_OWNER_LOGIN_TYPE env variable is not set",
```
johnstcn, line 61: `provider/workspace_owner.go:61:6: unknown field Summmary in struct literal of type diag.Diagnostic`

```suggestion
					Summary: "WARNING: The CODER_WORKSPACE_OWNER_LOGIN_TYPE env variable is not set",
```

johnstcn (approved):
Thanks for your contribution!

Note that this won't be useful until `coder/coder` is updated to set `CODER_WORKSPACE_OWNER_LOGIN_TYPE`. This will require a separate PR.

PR #324, opened 2024-12-20
chore: update paths to /v2

johnstcn:
We forgot to update the import paths when releasing v2! Doing so now as we'll probably need to import v2 of the provider in `coder/coder` eventually.


mtojek (approved):
Better late than never. Thanks!!

defelmnq (approved):

matifali:
Do we also need to cut a release to allow others importing this?

PR #327, opened 2025-01-20
fix: remove window option from open_in parameter

defelmnq:
The open_in parameter currently being released was first ment to contain `window` as one of the options - after more review it is not relevant anymore and has to be removed.


matifali (approved):
LGTM. Thanks

@johnstcn I am a bit unsure about versioning for this change.
Technically it is. breaking change.

integration/coder-app-open-in/main.tf
```diff
@@ -17,13 +17,6 @@ resource "coder_agent" "dev" {
   dir  = "/workspace"
 }
 
-resource "coder_app" "window" {
-  agent_id = coder_agent.dev.id
-  slug     = "window"
-  share    = "owner"
-  open_in  = "window"
-}
```
matifali, line 25: Should we change it to `tab` instead of removing it from the integration testing altogether?

integration/coder-app-open-in/main.tf
```diff
@@ -40,7 +33,6 @@ resource "coder_app" "defaulted" {
 locals {
   # NOTE: these must all be strings in the output
   output = {
-    "coder_app.window.open_in"      = tostring(coder_app.window.open_in)
     "coder_app.slim-window.open_in" = tostring(coder_app.slim-window.open_in)
```
mtojek, line 36: Don't you need to remove `slim-window` from here?

integration/coder-app-open-in/main.tf
```diff
@@ -17,18 +17,11 @@ resource "coder_agent" "dev" {
   dir  = "/workspace"
 }
 
-resource "coder_app" "window" {
+resource "coder_app" "tab" {
```
mtojek, line 20: The resource name is `tab`, so we the property `coder_app.tab.open_in` is available, but you're referring to `coder_app.slim-window.open_in` hence the issue?

defelmnq: Yes indeed , nice catch - just fixed it. 🤔

defelmnq:
Thanks @matifali @mtojek for the review and help - this one seems ready for me -

I'll need help as @matifali said above for the versioning - not sure what's the best here.

johnstcn:
> LGTM. Thanks
> 
> @johnstcn I am a bit unsure about versioning for this change. Technically, it is a breaking change. However, given that the added functionality in `2.1.0` was never released in Coder `mainline` or `stable`, it is safe to assume that no one should have been using it. And we can patch it as `2.1.1`. Thoughts?

Another option would be that we just make `window` default to `slim-window` and add a warning diagnostic. This would essentially have the same effect without being a 'breaking' change. 

WDYT @defelmnq ?

mtojek (approved):
I agree it is "breaking" but as long as nobody had a chance to use, maybe we should not leave "deprecation" notice? The notice will stay with us forever...

johnstcn (approved):
I'm OK with the current approach in any case. 👍

PR #331, opened 2025-01-27
feat: add resources_monitoring field to agent

defelmnq:
As required per [this issue](https://github.com/coder/internal/issues/245) - we need to add a new field in terraform-provider-coder , more specifically in the agent.


DanielleMaywood (approved):
LGTM :shipit:

examples/data-sources/coder_resources_monitoring/data-source.tf
```diff
@@ -0,0 +1,27 @@
+provider "coder" {}
+
+data "coder_provisioner" "dev" {}
+
+data "coder_workspace" "dev" {}
+
+resource "coder_agent" "main" {
+  arch = data.coder_provisioner.dev.arch
+  os   = data.coder_provisioner.dev.os
+  dir  = "/workspace"
```
dannykopping, line 10: I think this is out-of-date.

provider/agent.go
```diff
@@ -259,6 +259,66 @@ func agentResource() *schema.Resource {
...
+				Description: "The resources monitoring configuration for this agent.",
+				ForceNew:    true,
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"memory": {
+							Type:        schema.TypeSet,
+							Description: "The memory monitoring configuration for this agent.",
+							ForceNew:    true,
+							Optional:    true,
```
dannykopping, line 273: Consider adding `MinItems` and `MaxItems` for these resources.

provider/agent.go
```diff
@@ -259,6 +259,66 @@ func agentResource() *schema.Resource {
...
+										Type:        schema.TypeBool,
+										Description: "Enable volume monitoring for this agent.",
+										ForceNew:    true,
+										Required:    true,
+									},
+									"threshold": {
+										Type:        schema.TypeInt,
+										Description: "The volume usage threshold in percentage at which to trigger an alert. Value should be between 0 and 100.",
+										ForceNew:    true,
+										Required:    true,
```
dannykopping, line 314: Will we not have a sensible default?

provider/agent.go
```diff
@@ -259,6 +259,66 @@ func agentResource() *schema.Resource {
...
+										Type:        schema.TypeString,
+										Description: "The path of the volume to monitor.",
+										ForceNew:    true,
+										Required:    true,
+									},
+									"enabled": {
+										Type:        schema.TypeBool,
+										Description: "Enable volume monitoring for this agent.",
+										ForceNew:    true,
+										Required:    true,
```
dannykopping, line 308: If this resource is defined, it should be enabled by default.

provider/agent.go
```diff
@@ -259,6 +259,66 @@ func agentResource() *schema.Resource {
...
+						"volume": {
+							Type:        schema.TypeSet,
+							Description: "The volumes monitoring configuration for this agent.",
+							ForceNew:    true,
+							Optional:    true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"path": {
+										Type:        schema.TypeString,
+										Description: "The path of the volume to monitor.",
```
dannykopping, line 300: Validation of the volume can only be done in the agent
How do we plan to surface problems with accessing a volume - which may either be due to a permissions issue or an incorrect path?

provider/agent_test.go
```diff
@@ -211,6 +211,98 @@ func TestAgent_Metadata(t *testing.T) {
...
+				require.Equal(t, "/volume1", attr["resources_monitoring.0.volume.0.path"])
+				require.Equal(t, "100", attr["resources_monitoring.0.volume.1.threshold"])
+				require.Equal(t, "/volume2", attr["resources_monitoring.0.volume.1.path"])
+				return nil
+			},
+		}},
+	})
+}
+
+func TestAgent_ResourcesMonitoring_OnlyMemory(t *testing.T) {
```
dannykopping, line 267: Please add tests for unhappy cases like:
- multiple memory blocks
- volume blocks without paths defined
- thresholds outside accepted ranges
- ...etc

docs/resources/agent.md
```diff
@@ -116,3 +117,30 @@ Optional:
...
+
+
+<a id="nestedblock--resources_monitoring--volume"></a>
+### Nested Schema for `resources_monitoring.volume`
+
+Required:
+
+- `enabled` (Boolean) Enable volume monitoring for this agent.
+- `path` (String) The path of the volume to monitor.
+- `threshold` (Number) The volume usage threshold in percentage at which to trigger an alert. Value should be between 0 and 100.
```
matifali, line 146: This is just a suggestion on naming. The name `volume` seems provider-specific and mostly applies to pods and containers. I have also seen `disks` used for VMs.

What do you guys think about naming it `storage`?

dannykopping: "storage" is too generic.

The most logical common name (at least to me) is "volume"; it's a generic term used by both [Windows](https://support.microsoft.com/en-us/windows/disk-management-in-windows-ad88ba19-f0d3-0809-7889-830f63e94405) and [Linux](https://www.baeldung.com/linux/drive-partition-volumes).

I think it's pretty clear what "volume" means in all contexts, and we are designing it to be used with paths (which volumes are mounted at) it'll be a natural UX for operators.

provider/agent.go
```diff
@@ -259,6 +259,66 @@ func agentResource() *schema.Resource {
...
+							},
+						},
+						"volume": {
+							Type:        schema.TypeSet,
+							Description: "The volumes monitoring configuration for this agent.",
+							ForceNew:    true,
+							Optional:    true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"path": {
```
mtojek, line 298: We could do a simple validation in terraform-provider-coder:
* check if `path` is not just a string but an absolute path
* verify if `volume` set does not contain more entries with the same path (unless this is expected)

provider/agent.go
mtojek: +1

provider/agent.go
```diff
@@ -259,6 +259,66 @@ func agentResource() *schema.Resource {
...
+										ForceNew:    true,
+										Required:    true,
+									},
+									"enabled": {
+										Type:        schema.TypeBool,
+										Description: "Enable volume monitoring for this agent.",
+										ForceNew:    true,
+										Required:    true,
+									},
+									"threshold": {
```
mtojek, line 310: Validation please :)

examples/data-sources/coder_resources_monitoring/data-source.tf
mtojek: Can you modify `examples_test.go` to verify this file?

dannykopping (changes requested):
Blocking this from accidental merge since we already have one approval, but some work is still required for validation.

provider/agent.go
defelmnq: I added a `MaxItems` for memory to avoid duplicates. 👌

provider/agent.go
defelmnq: open to discussion here but I prefer fail than having a default as it could trigger the notifications - but both option can make sense to me.

provider/agent_test.go
defelmnq: I added some test cases which should improve a lot the coverage : 
- duplicate path
- multiple memory blocks
- thresholds out of range ( 0 - 100 )
and some OK tests.

provider/agent.go
defelmnq: I added some more tests and a better validation - should be way better now 🙏

defelmnq:
@mtojek @dannykopping - I improved the validation and added test cases for all the main cases, should be way better now. ✅

provider/agent.go
```diff
@@ -284,6 +288,12 @@ func agentResource() *schema.Resource {
 										Description: "The memory usage threshold in percentage at which to trigger an alert. Value should be between 0 and 100.",
 										ForceNew:    true,
 										Required:    true,
+										ValidateDiagFunc: func(i interface{}, s cty.Path) diag.Diagnostics {
+											if i.(int) < 0 || i.(int) > 100 {
+												return diag.Errorf("volume threshold must be between 0 and 100")
+											}
+											return nil
+										},
```
dannykopping, line 296: `github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation` has a bunch of built-in validations.

```suggestion
										ValidateDiagFunc: validation.IntBetween(0, 100),
```

provider/agent.go
dannykopping: OK 👍

provider/agent.go
```diff
@@ -321,29 +348,61 @@ func agentResource() *schema.Resource {
 			},
 		},
 		CustomizeDiff: func(ctx context.Context, rd *schema.ResourceDiff, i any) error {
-			if !rd.HasChange("metadata") {
-				return nil
+			if rd.HasChange("metadata") {
```
dannykopping, line 351: How did you test this code, and what was its output?

Also seems like we're just doing the validation here, which does not need seem to be the intent of this func: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/customizing-differences

provider/agent_test.go
```diff
@@ -241,65 +243,166 @@ func TestAgent_ResourcesMonitoring(t *testing.T) {
...
+	})
 
-				attr := resource.Primary.Attributes
-				require.Equal(t, "1", attr["resources_monitoring.#"])
-				require.Equal(t, "1", attr["resources_monitoring.0.memory.#"])
-				require.Equal(t, "80", attr["resources_monitoring.0.memory.0.threshold"])
-				return nil
-			},
-		}},
+	t.Run("NoPath", func(t *testing.T) {
```
dannykopping, line 384: These are good 👍 
You're missing too cases, though: empty volume and non-absolute path volume.

provider/agent.go
defelmnq: Indeed , for now the way it his done is :
- We validate that the path is not empty & this is an absolute one on tf-provider
- If the path is not accessible we ignore it on the agent side - but log it.

We can discuss the agent logic on the agent PR , but I feel like we do not want to block the worspace to start for monitoring , but definitely open to discuss it.

provider/agent.go
defelmnq: Yeah , this piece is maybe a bit weird. 

Globally, the metadata part was already validated here (at least for the duplicates) - I did not really changed the logic , just the if statement so we can have multiple cases. (so we check duplicates for both metadata and resources_monitoring).

About the reason why I've added the resources_monitoring here - the main one is because metadata was already validated here and the case was pretty much the same.

I tried to read the function comment and found :
```
CustomizeDiff is called after a difference (plan) has been generated
for the Resource and allows for customizations, such as setting values
not controlled by configuration, conditionally triggering resource
recreation, or implementing additional validation logic to abort a plan.
```

which seems to make sense to me , but still it's mostly a copy paste from previous logic.

defelmnq:
I added some more cases to ensure we test all new fields and options. 

Just a point remaining to discuss based on what we want to do around the duplicates validation.[ More details on this comment](https://github.com/coder/terraform-provider-coder/pull/331#discussion_r1945189190) but globally I copy pasted metadata - still if there's any better place I am open to change it.

Otherwise I tested the tf-provider with coder/coder and it works exactly as we want - so I'm happy about the current behavior.

provider/agent.go
dannykopping: > I feel like we do not want to block the worspace to start for monitoring , but definitely open to discuss it.

Do you mean that if the resource monitoring definition is somehow wrong it should still start the agent but log an error/warning?

provider/agent.go
```diff
@@ -259,31 +261,138 @@ func agentResource() *schema.Resource {
...
+							Optional:    true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"path": {
+										Type:        schema.TypeString,
+										Description: "The path of the volume to monitor.",
+										ForceNew:    true,
+										Required:    true,
+										ValidateDiagFunc: func(i interface{}, s cty.Path) diag.Diagnostics {
+											if i.(string) == "" {
```
dannykopping, line 309: This will probably be fine, but in general you should always check that a type-cast has succeeded, otherwise you can panic the provider.

provider/agent.go
dannykopping: OK. There's a part of my question still unanswered though:

> How did you test this code, and what was its output?

provider/agent_test.go
```diff
@@ -211,6 +211,302 @@ func TestAgent_Metadata(t *testing.T) {
...
+						arch = "amd64"
+						resources_monitoring {
+							memory {
+								enabled = true
+								threshold = 101
+							}
+						}
+					}`,
+				Check:       nil,
+				ExpectError: regexp.MustCompile("Error running pre-apply refresh"),
```
dannykopping, line 350: Huh? This doesn't seem like the validation error I would expect.

provider/agent_test.go
dannykopping: Nice 👍 thanks

matifali:
I have two suggestions.
1. Can we add an error that provider emits if someone tries to use `resource_monitoring` with Coder version 2.19 and below? The Error should tell the user what minimum Coder version the provider needs for this feature to work.
2. Tag this release as `v2.2.0`

provider/agent.go
defelmnq: Oh yeah sorry - for this part the code is tested with the duplicates. 
Metadata (was already done) - there's a test `TestAgent_MetadataDuplicateKeys` 
Resources monitoring - `DuplicatePaths`

provider/agent.go
defelmnq: Sorry maybe was not precise enough - yes, but only for the situation on the which terraform itself is correct, validation is done BUT the volume is not present in the workspace, or not accessible.

For example `/user/coder` is a valid absolute path, but maybe do not exist in a given workspace - in that situation I would do what describe above - start the workspace & agent, but display a warning.

dannykopping: Sounds good 👍

provider/agent.go
dannykopping: Ah cool 👍 thanks

provider/agent_test.go
defelmnq: My bad, used this one as I did not found a way to make the regexp work, a bit more work and it is now good - should be more explicit 👌

provider/agent.go
defelmnq: Better safe than sorry - you're completely right so I added a small check.

defelmnq:
I'll indeed defer answers to your questions to @johnstcn - not sure if this is doable and how , but happy to implement the check if we want to.

About the versioning, both are fine to me. I already tested it a lot with the Coder code we will deploy but a rc can be good too.

dannykopping (approved):
LGTM!

mtojek (approved):
Nice!

johnstcn:
> I have two suggestions.
> 
> 1. Can we add an error that provider emits if someone tries to use `resource_monitoring` with Coder version 2.19 and below? The Error should tell the user what minimum Coder version the provider needs for this feature to work.

It should be possible to hit the `/api/v2/buildinfo` endpoint. We can't import `codersdk` here though so we'd need to do a little bit of copying. But it should definitely be possible! I think that's a good follow-up PR.

> 2. Tag the upcoming release as `v2.2.0`. We can also use `v2.2.0-rc.0` until we release a compatible Coder release and then move to `v2.2.0`.
>    > _One downside of an RC release is that if we fix anything else, we need to do a patch from `main` that will include this new feature. I also do not want us to release from branches._
> 
> WDYT @johnstcn as we recently spoke about versions for our provider?

I'm fine with RC releases. It's a little annoying to have to cherry-pick each time, but it's not the end of the world.

johnstcn (approved):

PR #332, opened 2025-01-28
feat: add a workspace preset datasource

SasSwart:
This pull request adds `data "coder_workspace_preset" {}`

Basic validation is done when the data source is read, but cross validation against `coder_parameter` data sources happens in the provisioner in coder/coder.

Once merged into main, we'll need to backport to `v1.0.5` and bump the dependency in `coder/coder`, because we rely on types in this provider in the coder provisioner.


provider/workspace_preset.go
```diff
@@ -0,0 +1,39 @@
...
+
+func workspacePresetDataSource() *schema.Resource {
+	return &schema.Resource{
+		SchemaVersion: 1,
+
+		Description: "",
+		ReadContext: func(ctx context.Context, rd *schema.ResourceData, i interface{}) diag.Diagnostics {
+			rd.SetId(uuid.NewString())
+
+			return nil
```
SasSwart, line 24: TODO: 
double check whether I'm missing anything here. 
If not, comment why its implementation is this bare.
Also check the purpose and correct value for `rd.SetId(...)`

provider/workspace_preset_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWorkspacePreset(t *testing.T) {
+	resource.Test(t, resource.TestCase{
```
SasSwart, line 12: This looks like the right place to add tests for parameter validation.

SasSwart (commented):
Self review:

provider/workspace_preset.go
```diff
@@ -0,0 +1,64 @@
...
+			}, &preset)
+			if err != nil {
+				return diag.Errorf("decode workspace preset: %s", err)
+			}
+
+			if preset.Name == "" {
+				return diag.Errorf("workspace preset name must be set")
+			}
+
+			if len(preset.Parameters) == 0 {
```
dannykopping, line 38: This should be implemented as validation options in the schema.

See `MinItems` for `parameters`.

Interesting/weird that the validation doesn't catch empty strings with its `Required` attribute. Thanks for checking both 👍

provider/workspace_preset.go
```diff
@@ -0,0 +1,64 @@
...
+				return diag.Errorf("workspace preset must define a value for at least one parameter")
+			}
+
+			rd.SetId(preset.Name)
+
+			return nil
+		},
+		Schema: map[string]*schema.Schema{
```
dannykopping, line 46: In terms of additional validation, I don't _think_ it's possible to access the state of other datasources (i.e. `coder_parameter`s) to ensure the parameters are set, but curious if you looked into this.

provider/workspace_preset_test.go
```diff
@@ -0,0 +1,127 @@
...
+			Config: `
+			data "coder_workspace_preset" "preset_1" {
+				name = [1, 2, 3]
+				parameters = {
+					"region" = "us-east1-a"
+				}
+			}`,
+			ExpectError: regexp.MustCompile("Incorrect attribute value type"),
+		}},
+	})
```
dannykopping, line 84: Nit: appreciate the completeness here but this is testing Terraform's validation, which seems unnecessary.

provider/workspace_preset_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Steps: []resource.TestStep{{
+			Config: `
+			data "coder_workspace_preset" "preset_1" {
+				name = "preset_1"
+				parameters = "not a map"
+			}`,
+			// This is from terraform's validation based on our schema, not based on our validation in ReadContext:
+			ExpectError: regexp.MustCompile("Inappropriate value for attribute \"parameters\": map of string required"),
+		}},
+	})
```
dannykopping, line 126: Nit: ditto.

dannykopping (approved):
Minor comment about the idiomatic approach to validation but otherwise LGTM
Feel free to ignore the nits.

provider/workspace_preset.go
SasSwart: MinItems doesn't seem to work with Maps. Only with lists or sets. It's the one part of custom validation that I've needed to retain. I've been able to remove all the rest in favour of adding the the Schema though.

provider/workspace_preset.go
SasSwart: It's not possible out of the box using the SDK, but there are two ways that I have thought of to validate against `coder_parameter` validations. I still need to prototype them but that will be for a separate PR.

provider/workspace_preset_test.go
SasSwart: I've added comments to explain why I'm doing it. TL;DR: While terraform is technically the tool doing the validation, we still need to correctly configure the schema. Regardless of what generates the error message, I'm testing the behaviour.

PR #339, opened 2025-02-07
fix(docs): fix parameter options icons

### LINKED ISSUE #336: coder_parameter doc has examples with country icons that don't exist
https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/parameter cites:

```
data "coder_parameter" "example" {
  name        = "Region"
  description = "Specify a region to place your workspace."
  mutable     = false
  type        = "string"
  default     = "asia-central1-a"
  option {
    value = "us-central1-a"
    name  = "US Central"
    icon  = "/icon/usa.svg"
  }
  option {
    value = "asia-central1-a"
    name  = "Asia"
    icon  = "/icon/asia.svg"
  }
}
```

These country icons do not currently exist in 2.18.4. In fact, when you go to a URL like https://coder.example.com/icon/usa.svg, you'll get a 200 with no body that does some XHR that returns this back to the view:

Image

matifali:
Resolves #336

### DIFF
--- a/docs/data-sources/parameter.md
+++ b/docs/data-sources/parameter.md
@@ -20,7 +20,7 @@ data "coder_parameter" "example" {
   description = "Specify a region to place your workspace."
   mutable     = false
   type        = "string"
-  default     = "asia-central1-a"
+  default     = "us-central1-a"
   option {
     value = "us-central1-a"
     name  = "US Central"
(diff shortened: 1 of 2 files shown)


mtojek (approved):
quick stamp 👍

johnstcn (approved):

PR #340, opened 2025-02-07
ci: drop support for EOL terraform version

matifali:
Drop support for EOL terraform versions. 
We upgraded to terraform 1.9 in Coder 2.14.x 7 months ago, which is no longer supported.

Also includes terraform 1.10


johnstcn (approved):
:shipit:

PR #347, opened 2025-02-19
chore: test auto prerelease

matifali:
This configures goreleaser to automate prereleases when we push a tag that contains `pre` or `rc`
Ref: https://goreleaser.com/customization/release/#github

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -56,6 +56,8 @@ release:
       name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
   # If you want to manually examine the release before its live, uncomment this line:
   # draft: true
+  prerelease: auto
+  make_latest: "{{ not .Prerelease }}"
 changelog:
   # see https://goreleaser.com/customization/changelog/
   use: github-native
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -57,7 +57,7 @@ release:
   # If you want to manually examine the release before its live, uncomment this line:
   # draft: true
   prerelease: auto
-  make_latest: "{{ not .Prerelease }}"
+  make_latest: {{ eq (.Prerelease | len) 0 }}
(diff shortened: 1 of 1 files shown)


johnstcn (approved):
🤞

.goreleaser.yml
```diff
@@ -56,6 +56,8 @@ release:
       name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
   # If you want to manually examine the release before its live, uncomment this line:
   # draft: true
+  prerelease: auto
+  make_latest: "{{ not .Prerelease }}"
```
matifali, line 60: As per https://goreleaser.com/customization/templates/

```yaml
make_latest: "{{ not .Prerelease }}"
```

is wrong and

```yaml
"{{ .Prerelease }}"
```
will evaluate to `pre`
It is not a boolean, so my change will not work.

QQ: Can I match a template check for
```yaml
"{{ .Prerelease == 'pre' }}"
```

I am assuming templating works the same as they do in Helm charts.

johnstcn: It looks like templates aren't supported in `make_latest` yet. 
There's a PR to support this: https://github.com/goreleaser/goreleaser/issues/5418

matifali: I think it's merged and available as per v2.6.0
https://github.com/goreleaser/goreleaser/pull/5420

johnstcn: `{{ eq (.Prerelease | len) 0 }}`

https://go.dev/play/p/4_PNM98L86r

johnstcn (approved):
`{{ not .Prerelease }}` should also work as you suggested: https://go.dev/play/p/aDU6YypI2KT

.goreleaser.yml
matifali: Reverting the change.

PR #354, opened 2025-02-28
DNM: hack: terraform-plugin-framework and dynamic types

johnstcn:


Emyrk:
I see type constraint values might still not be supported? So 

```hcl
data "coder_parameter" "test" {
  type = string // Not allowed
  type = "string" // is allowed
```

It treats `string` as a reference, rather than a type constraint. This does check out though, as type constraints are treated as keywords. And keywords in expressions are just traversals.

See traversal to keyword: https://github.com/hashicorp/hcl/blob/main/traversal_for_expr.go#L70-L127
And how they parse the keyword to a type: https://github.com/hashicorp/terraform/blob/main/internal/stacks/stackconfig/typeexpr/typeexpr.go#L27-L31

And on the input value: https://github.com/hashicorp/terraform/blob/main/internal/stacks/stackconfig/input_variable.go#L20-L21 (var blocks are input variables)


So type constraints are just references. We might have to keep our types as quoted strings? :cry:

johnstcn:
> Traversals less than 2 are considered invalid too? 

I think that's just for a resource, which is always `resource_type.name`. See `parseRef` for other kinds of traversals https://github.com/terraform-linters/tflint-plugin-sdk/blob/master/terraform/addrs/parse_ref.go#L66

Emyrk:
Another issue with the current sdk. 

https://github.com/coder/terraform-provider-coder/pull/123

The current does not support pointer types? So we have to compute some helper fields?

Emyrk:
another reason to do this: https://github.com/coder/preview/issues/40

PR #355, opened 2025-03-05
chore: do not use GitHub prereleases

matifali:
Removes automatic handling of GitHub releases.
For some reason, the registry isn't recognizing new tags, and we suspect this might not be the cause.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -55,10 +55,6 @@ release:
   extra_files:
     - glob: 'terraform-registry-manifest.json'
       name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
-  # If you want to manually examine the release before its live, uncomment this line:
-  # draft: true
-  prerelease: auto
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -55,10 +55,6 @@ release:
   extra_files:
     - glob: 'terraform-registry-manifest.json'
       name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
-  # If you want to manually examine the release before its live, uncomment this line:
-  # draft: true
-  prerelease: auto
-  make_latest: '{{ not .Prerelease }}'
```
matifali, line 61: default values are `prerelease: false` and `make_latest: true`
https://goreleaser.com/customization/release/#github

dannykopping (approved):

johnstcn (commented):
😭

matifali:
We discovered that the Terraform registry does not pick up them, although we could not find any documentation.

PR #357, opened 2025-03-05
chore: only release if tests pass

dannykopping:
We've generally been lucky that builds are slower than integration tests, but we need explicit ordering here


matifali (approved):

johnstcn (commented):
Post-merge 👍

PR #330, opened 2025-01-23
feat: add `rbac_roles` to `coder_workspace_owner` data source

nxf5025:
This PR updates the schema for the coder_workspace_owner data source to expose rbac_roles.

It's very similar to https://github.com/coder/terraform-provider-coder/pull/287

Relevant `coder/coder` PR - https://github.com/coder/coder/pull/16407

PR #369, opened 2025-03-19
chore(integration): fix failing tests referencing workspace_owner.rbac_roles

johnstcn:
The tests updated in #330 may have worked initially when run against `coder-preview:latest` but that's a moving target.

Updates existing test to make no assertions (empty regexp) and adds a new test specifically for rbac roles field. Yes, it's a janky regex.
Ideally we'd have a way to:
1) Conditionally assert specific fields of resources.
2) Do JSON-ish assertions.

Closes #367


integration/workspace-owner-rbac-roles/main.tf
```diff
@@ -0,0 +1,59 @@
...
+    coder = {
+      source = "coder/coder"
+    }
+    local = {
+      source = "hashicorp/local"
+    }
+  }
+}
+
+// TODO: test coder_external_auth
```
mafredri, line 12: Copy-pasted TODO, to-don't?

mafredri (approved):
Looks OK to me, although I didn't spend a lot of time verifying it. Thanks for fixing 👍🏻

integration/workspace-owner-rbac-roles/main.tf
johnstcn: That is the question!

PR #373, opened 2025-04-03
feat: allow presets to define prebuilds

SasSwart:
This PR closes https://github.com/coder/terraform-provider-coder/issues/362

### DIFF
--- a/provider/workspace.go
+++ b/provider/workspace.go
@@ -91,20 +91,15 @@ func workspaceDataSource() *schema.Resource {
 				Computed:    true,
 				Description: "The access port of the Coder deployment provisioning this workspace.",
 			},
+			"prebuild_count": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "A computed count, equal to 1 if the workspace was prebuilt.",
+			},
 			"start_count": {
 				Type:        schema.TypeInt,
 				Computed:    true,
 				Description: "A computed count based on `transition` state. If `start`, count will equal 1.",
-			},
-			"prebuild_count": {
-				Type:        schema.TypeInt,
-				Computed:    true,
-				Description: "TODO",
-			},
-			"is_prebuild": {
-				Type:        schema.TypeBool,
-				Computed:    true,
-				Description: "TODO",
 			},
 			"transition": {
 				Type:        schema.TypeString,
@@ -115,6 +110,11 @@ func workspaceDataSource() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 				Description: "UUID of the workspace.",
+			},
+			"is_prebuild": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether the workspace is a prebuild.",
 			},
 			"name": {
 				Type:        schema.TypeString,
--- a/provider/workspace.go
+++ b/provider/workspace.go
@@ -27,6 +27,14 @@ func workspaceDataSource() *schema.Resource {
 			}
 			_ = rd.Set("start_count", count)
 
+			prebuild := helpers.OptionalEnv(IsPrebuildEnvironmentVariable())
+			prebuildCount := 0
+			if prebuild == "true" {
+				prebuildCount = 1
+				_ = rd.Set("is_prebuild", true)
+			}
+			_ = rd.Set("prebuild_count", prebuildCount)
+
 			name := helpers.OptionalEnvOrDefault("CODER_WORKSPACE_NAME", "default")
 			rd.Set("name", name)
 
@@ -88,6 +96,16 @@ func workspaceDataSource() *schema.Resource {
 				Computed:    true,
 				Description: "A computed count based on `transition` state. If `start`, count will equal 1.",
 			},
+			"prebuild_count": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "TODO",
+			},
+			"is_prebuild": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "TODO",
+			},
 			"transition": {
 				Type:        schema.TypeString,
 				Computed:    true,
@@ -120,4 +138,8 @@ func workspaceDataSource() *schema.Resource {
 			},
 		},
 	}
+}
+
+func IsPrebuildEnvironmentVariable() string {
+	return "CODER_WORKSPACE_IS_PREBUILD"
 }
--- a/provider/workspace.go
+++ b/provider/workspace.go
@@ -94,7 +94,7 @@ func workspaceDataSource() *schema.Resource {
 			"prebuild_count": {
 				Type:        schema.TypeInt,
 				Computed:    true,
-				Description: "A computed count, equal to 1 if the workspace was prebuilt.",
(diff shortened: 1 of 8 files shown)


provider/workspace.go
```diff
@@ -83,6 +91,11 @@ func workspaceDataSource() *schema.Resource {
 				Computed:    true,
 				Description: "The access port of the Coder deployment provisioning this workspace.",
 			},
+			"prebuild_count": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "A computed count, equal to 1 if the workspace was prebuilt.",
```
mafredri, line 97: _was prebuilt_, I may be wrong but could we improve terminology here? If my hunch is right both of these are true: "is a prebuild", "was prebuilt"?

provider/workspace.go
```diff
@@ -98,6 +111,11 @@ func workspaceDataSource() *schema.Resource {
 				Computed:    true,
 				Description: "UUID of the workspace.",
 			},
+			"is_prebuild": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether the workspace is a prebuild.",
```
mafredri, line 117:
```suggestion
				Description: "Set to true if the workspace is a prebuild.",
```
Pet-peeve: Whether doesn't describe which is the expected value, even if it can be inferred from `is_`.

provider/workspace_preset.go
```diff
@@ -65,6 +80,22 @@ func workspacePresetDataSource() *schema.Resource {
 					ValidateFunc: validation.StringIsNotEmpty,
 				},
 			},
+			"prebuilds": {
+				Type:        schema.TypeSet,
+				Description: "Prebuilds of the workspace preset.",
```
mafredri, line 85: I think I have a unique view as I lack much context around prebuilds and presets, but to me (as a user) this doesn't say much. It could potentially explain how they tie together or how setting this changes something.

mafredri (approved):
I think the docs could be improved a bit but otherwise LGTM 👍🏻

johnstcn (commented):
At the risk of opening a can of worms, I think it would be valuable to add or modify an existing integration test for this -- even if it just asserts no prebuilds exist right now.

provider/workspace_preset.go
```diff
@@ -10,8 +10,13 @@ import (
 )
 
 type WorkspacePreset struct {
-	Name       string            `mapstructure:"name"`
-	Parameters map[string]string `mapstructure:"parameters"`
+	Name       string              `mapstructure:"name"`
+	Parameters map[string]string   `mapstructure:"parameters"`
+	Prebuild   []WorkspacePrebuild `mapstructure:"prebuilds"`
```
evgeniy-scherbina, line 15: Why we use `[]WorkspacePrebuild` instead of `WorkspacePrebuild`?

provider/workspace.go
```diff
@@ -83,6 +91,11 @@ func workspaceDataSource() *schema.Resource {
 				Computed:    true,
 				Description: "The access port of the Coder deployment provisioning this workspace.",
 			},
+			"prebuild_count": {
```
evgeniy-scherbina, line 94: `is_prebuild` & `prebuild_count` looks the same to me, what's the point to have both?

provider/workspace_preset.go
SasSwart: Your lack of context is valuable. You already have more context than a hypothetical new user. If it's vague to you, it will be even more confusing to users. So we should definitely improve it. Will do.

provider/workspace_preset.go
SasSwart: I couldn't find a reason for this. I've changed it to just WorkspacePrebuild.

provider/workspace.go
SasSwart: This seems to be duplication, yes. We could probably get rid of "is_prebuild" and just check the count. Looking at the rest of the code, we are following the pattern that was set by the "transition" and "start_count" parameters. They have the same relationship.

I'm not sure whether to remove "is_prebuild" or keep it.

provider/workspace.go
SasSwart: I've updated the documentation.

provider/workspace_preset.go
SasSwart: I've added additional documentation.

provider/workspace.go
evgeniy-scherbina: I don't like neither `prebuild_count` nor `start_count`. But probably there is some reason we're using, I'm not aware of.
My guess is we're using `start_count` to map transition statuses, like:
- stop: 0
- start: 1
etc...

Also looks like there is some duplication here as well:
- start_count
- transition

But I don't know why it's needed.

We can ask original author of this approach with `start_count`, `transition`?
Maybe it was done in a rush, and there is no sense to continue using this approach.

provider/workspace.go
```diff
@@ -27,6 +27,14 @@ func workspaceDataSource() *schema.Resource {
 			}
 			_ = rd.Set("start_count", count)
 
+			prebuild := helpers.OptionalEnv(IsPrebuildEnvironmentVariable())
+			prebuildCount := 0
+			if prebuild == "true" {
+				prebuildCount = 1
+				_ = rd.Set("is_prebuild", true)
+			}
+			_ = rd.Set("prebuild_count", prebuildCount)
```
evgeniy-scherbina, line 36: nit: I'd rather write it like this:
```go
if prebuild == "true" {
	_ = rd.Set("is_prebuild", true)
	_ = rd.Set("prebuild_count", 1)
} else {
	_ = rd.Set("is_prebuild", false)
	_ = rd.Set("prebuild_count", 0)
}
```

but it's not a big deal

provider/workspace.go
```diff
@@ -121,3 +139,7 @@ func workspaceDataSource() *schema.Resource {
 		},
 	}
 }
+
+func IsPrebuildEnvironmentVariable() string {
+	return "CODER_WORKSPACE_IS_PREBUILD"
```
evgeniy-scherbina, line 144: I guess it can be const instead of func, but up to you

provider/workspace_preset.go
```diff
@@ -12,59 +12,82 @@ import (
...
 }
 
 func workspacePresetDataSource() *schema.Resource {
 	return &schema.Resource{
 		SchemaVersion: 1,
 
-		Description: "Use this data source to predefine common configurations for workspaces.",
+		Description: "Use this data source to predefine common configurations for coder workspaces. Users will have the option to select a defined preset, which will automatically apply the selected configuration. Any parameters defined in the preset will be applied to the workspace. Parameters that are not defined by the preset will still be configurable when creating a workspace.",
 		ReadContext: func(ctx context.Context, rd *schema.ResourceData, i interface{}) diag.Diagnostics {
 			var preset WorkspacePreset
```
evgeniy-scherbina, line 28: @SasSwart btw: I just noticed code in the `ReadContext` is basically no-op?
It reads data and then forgets them.
Probably it's fine if main goal it to send this data downstream to `provisioner`

But still what is the purpose of this code?
- is it for validation purposes? But I'm not sure maybe validations happens earlier?
- is it for helping debugging in the future?

SasSwart: Most of the decoding is defunct, yes. I've removed it.
I don't think we should rely on this for validation.
We define a schema below, which Terraform already uses for validation.
I've removed the defunct code here: https://github.com/coder/terraform-provider-coder/pull/376

Issue #364, opened 2025-03-17
Ensure the workspace agent can reuse its existing token when required


SasSwart:
In `coderd`, when we provision a prebuild, we start a normal running workspace. This workspace has an agent running on it, and the design of prebuilds relies on neither the workspace nor the agent having to restart when the prebuild is claimed (if they restart, this obviates the prebuild because the start time can be very slow).

When a workspace has a `coder_agent` defined in its template, this agent resource will be created in the `workspace_agents` table upon `terraform apply`. All other template collateral, including the VM/container/whatever on which the agent runs, will be recorded in the `workspace_resources` table. The agent has an FK association to the `workspace_resources` entry of the compute on which it runs.

When the `coder_agent` resource is created in this provider, [it generates itself a new token](https://github.com/coder/terraform-provider-coder/blob/1f95807/provider/agent.go#L28). This token is persisted to the database in https://github.com/coder/coder/blob/9469b78290674238e1d94cdee41a5cfec13374ec/provisioner/terraform/resources.go#L338-L342

When a prebuild is claimed, we run another `terraform apply` to effect any desired changes.

It should be noted that the VM and the agent will **ONLY** be created in the provider (i.e. AWS) on the first `terraform apply` (when the prebuild is created), and the VM should **ONLY** be modified in-place on subsequent `apply`es.

After the second apply, the control plane will then issue an instruction for the agent to "reinitialize". This effectively just means fetching a new manifest from the control plane, which will include the new ownership information. As stated above, the actual VM or running agent process will not be recreated, and therefore the running agent needs to retain its token that it was initially launched with.

We should pass down the running agent's token to the provider to replace the autogenerated token, for this purpose.

The subsequently created `workspace_agents` record will contain the auth token of the initial build (i.e. prebuild creation), and this will mean that the agent's authentication will succeed via https://github.com/coder/coder/blob/5841c0a/coderd/httpmw/workspaceagent.go#L60.

![Image](https://github.com/user-attachments/assets/e9127acb-db2e-4ebd-bee7-49d1ae80cf23)

PR #374, opened 2025-04-03
feat: reuse agent tokens when a prebuilt agent reinitializes

SasSwart:
This PR closes https://github.com/coder/terraform-provider-coder/issues/364


johnstcn (commented):
I think this is _definitely_ worth an integration test :-)

dannykopping (commented):
I don't think this PR is ready for review yet, is it?

provider/agent.go
```diff
@@ -22,10 +25,38 @@ func agentResource() *schema.Resource {
...
+			// existingToken should only have been set if isPrebuild is true, because we only
+			// reuse the token when a prebuilt workspace is being claimed.
+			existingToken := helpers.OptionalEnv(RunningAgentTokenEnvironmentVariable(agentID))
+			logFields := map[string]interface{}{
+				"agent_id":       agentID,
+				"is_prebuild":    isPrebuild,
+				"token_provided": existingToken != "",
+			}
+			if isPrebuild && existingToken != "" {
+				// check if a token was already generated for this agent.
```
dannykopping, line 50: Nit: consistency.

```suggestion
				// Check if a token was already generated for this agent.
```

provider/agent.go
```diff
@@ -22,10 +25,38 @@ func agentResource() *schema.Resource {
...
+				// If so, this workspace is in the process of being claimed
+				// and we should reuse the token. If not, we use a new token as usual.
+				tflog.Info(ctx, "using provided agent token for prebuild", logFields)
+				token = existingToken
+			} else {
+				tflog.Info(ctx, "using a new agent token", logFields)
+			}
```
dannykopping, line 57: I don't think it's necessary to log this case.

```suggestion
			}
```

provider/agent.go
```diff
@@ -469,3 +500,16 @@ func updateInitScript(resourceData *schema.ResourceData, i interface{}) diag.Dia
 	}
 	return nil
 }
+
+// RunningAgentTokenEnvironmentVariable returns the name of the environment variable
+// that contains the token for the running agent. This is used for prebuilds, where
+// we want to reuse the same token for the next iteration of a workspace agent before
+// and after the workspace was claimed by a user.
```
dannykopping, line 507: This comment could use some work; I think we should elaborate as to why this is being injected.

provider/agent.go
```diff
@@ -469,3 +500,16 @@ func updateInitScript(resourceData *schema.ResourceData, i interface{}) diag.Dia
...
+//
+// agentID is unused for now, but will be used as soon as we support multiple agents.
+func RunningAgentTokenEnvironmentVariable(agentID string) string {
```
dannykopping, line 510: How will the ID be injected? How will we maintain a persistent identity across `terraform apply` runs?

dannykopping (approved):

PR #375, opened 2025-04-03
feat: `form_type` and `styling` metadata arguments added

Emyrk:
This PR is in support of dynamic parameters. Changes to coder/coder intends to support this experimentally by the end of April. For current versions of coder, these arguments will be ignored.

## What is this?

This adds `form_type` as a field to parameters. By default, legacy behavior is maintained by deducing the default `form_type` value from the existing inputs.

## Why do this?

To support more form type options in the UI. See the truth table here of all the new types:

https://github.com/coder/terraform-provider-coder/blob/2b5ff0815acbb57fdca704b7123f9fa3722ab100/provider/formtype.go#L74-L91

## Example new param

```terraform
data "coder_parameter" "tools" {
  name        = "Tools"
  description = "What tools do you want to install?"
  type        = "list(string)"
  form_type   = "multi-select"
  styling = jsonencode({
    # Arbitrary JSON object to be passed to the frontend
  })
  default = jsonencode(["git", "docker"])

  option {
    value = "Docker"
    name  = "docker"
    icon  = "/icon/docker.svg"
  }
  option {
    value = "Git"
    name  = "git"
    icon  = "/icon/git.svg"
  }
  option {
    value = "Golang"
    name  = "go"
    icon  = "/icon/go.svg"
  }
  option {
    value = "Typescript"
    name  = "ts"
    icon  = "/icon/typescript.svg"
  }
}
```

## Note on docs

I am unable to add more example params to the docs, as they are used in integration tests that seem to use provider versions without this change? I get this error for above:

```
2025-04-07T18:26:57.1664097Z     examples_test.go:30: Step 1/1 error: Error running pre-apply refresh: exit status 1
2025-04-07T18:26:57.1664830Z         
2025-04-07T18:26:57.1665915Z         Error: default value "[\"git\",\"docker\"]" is not a valid option, values "git, docker" are missing from the option
2025-04-07T18:26:57.1666653Z         
2025-04-07T18:26:57.1667135Z           with data.coder_parameter.tools,
2025-04-07T18:26:57.1668078Z           on terraform_plugin_test.tf line 119, in data "coder_parameter" "tools":
2025-04-07T18:26:57.1668889Z          119: data "coder_parameter" "tools" {
```

That error is the error you get using this param on the **existing** terraform provider. Maybe I have to do some skips with past versions. Regardless, I will update the docs when this gets further upstream to a release.


provider/formtype_test.go
```diff
@@ -0,0 +1,426 @@
...
+					options:    true,
+				})
+			}
+		}
+
+		for _, check := range requiredChecks {
+			_, alreadyChecked := formTypesChecked[check.String()]
+			if alreadyChecked {
+				continue
+			}
```
johnstcn, line 288:
```suggestion
			if _, alreadyChecked := formTypesChecked[check.String()]; alreadyChecked {
				continue
			}
```

provider/formtype_test.go
```diff
@@ -0,0 +1,426 @@
...
+						Options:    %t,
+						OptionType: %q,
+						FormType:   %q,
+					}),
+				//`, "<expected_form_type>", check.options, check.optionType, check.formType)
+
+				probablyPassed := formTypeTest(t, fc)
+				if !probablyPassed {
+					t.Logf("To construct this test case:\n%s", tcText)
+				}
```
johnstcn, line 321: What does `probablyPassed` mean here?

provider/formtype_test.go
```diff
@@ -0,0 +1,426 @@
...
+			})
+
+		}
+	})
+}
+
+// ezconfig converts a formTypeCheck into a terraform config string.
+func ezconfig(paramName string, cfg formTypeCheck) (defaultValue string, tf string) {
```
johnstcn, line 329: nit: rename for clarity, `ezconfig` isn't particularly illuminating as an identifier.

provider/formtype_test.go
```diff
@@ -0,0 +1,426 @@
...
+		require.Len(t, state.Modules, 1)
+		require.Len(t, state.Modules[0].Resources, 1)
+
+		key := strings.Join([]string{"data", "coder_parameter", paramName}, ".")
+		param := state.Modules[0].Resources[key]
+
+		probablyPassed = probablyPassed && assert.Equal(t, def, param.Primary.Attributes["default"], "default value")
+		probablyPassed = probablyPassed && assert.Equal(t, string(c.assert.FormType), param.Primary.Attributes["form_type"], "form_type")
+		probablyPassed = probablyPassed && assert.Equal(t, string(c.assert.Type), param.Primary.Attributes["type"], "type")
+		probablyPassed = probablyPassed && assert.JSONEq(t, string(c.assert.Styling), param.Primary.Attributes["styling"], "styling")
```
johnstcn, line 402: Suggestion: rewrite the assertion chain and use a buffered channel to pass the result back.
```
resultCh <- assert.Equal(cond1) && assert.Equal(cond2) && ...
```

provider/formtype_test.go
```diff
@@ -0,0 +1,426 @@
...
+					%s
+				}
+		`, paramName, paramName, body.String())
+}
+
+func formTypeTest(t *testing.T, c formTypeTestCase) bool {
+	t.Helper()
+	const paramName = "test_param"
+	// probablyPassed is just a guess used for logging. It's not important.
+	probablyPassed := true
```
johnstcn, line 389: One the one hand, you say that it's not important, but it's returned by the function here. This implies to me that it is arbitrarily as important as whatever callers decide to do with it?

provider/parameter.go
```diff
@@ -170,9 +187,36 @@ func parameterDataSource() *schema.Resource {
...
-						return diag.Errorf("default value %q must be defined as one of options", parameter.Default)
+					if parameter.Type == OptionTypeListString && optionType == OptionTypeString {
+						// If the type is list(string) and optionType is string, we have
+						// to ensure all elements of the default exist as options.
+						var defaultValues []string
+						// TODO: We do this unmarshal in a few spots. It should be standardized.
+						err = json.Unmarshal([]byte(parameter.Default), &defaultValues)
+						if err != nil {
+							return diag.Errorf("default value %q is not a list of strings", parameter.Default)
+						}
```
johnstcn, line 198: Good spot! 👍

provider/parameter.go
```diff
@@ -170,9 +187,36 @@ func parameterDataSource() *schema.Resource {
...
+
+						// missing is used to construct a more helpful error message
+						var missing []string
+						for _, defaultValue := range defaultValues {
+							_, defaultIsValid := values[defaultValue]
+							if !defaultIsValid {
+								missing = append(missing, defaultValue)
+							}
```
johnstcn, line 206: suggestion: combine condition and assignment

provider/parameter.go
```diff
@@ -170,9 +187,36 @@ func parameterDataSource() *schema.Resource {
 				}
 
 				if parameter.Default != "" {
-					_, defaultIsValid := values[parameter.Default]
-					if !defaultIsValid {
-						return diag.Errorf("default value %q must be defined as one of options", parameter.Default)
+					if parameter.Type == OptionTypeListString && optionType == OptionTypeString {
```
johnstcn, line 190: This feels ripe to be extracted to a function.

provider/parameter_test.go
```diff
@@ -686,9 +689,9 @@ data "coder_parameter" "region" {
...
 	for _, tc := range []struct {
-		Name,
-		Type,
-		Value,
+		Name  string
+		Type  provider.OptionType
+		Value string
 		Regex,
 		RegexError string
 		Min,
```
johnstcn, line 697: TIL you can elide these. If we re-arrange the struct fields does that mean we can just elide the type of `RegexError` as well?

e.g.

```
Type provider.OptionType
Name string
Value,
Regex,
RegexError,
Min, ...
```

provider/formtype_test.go
```diff
@@ -0,0 +1,426 @@
...
+
+	expectType := func(expected provider.ParameterFormType, opts formTypeCheck) formTypeTestCase {
+		ftname := opts.formType
+		if ftname == "" {
+			ftname = "default"
+		}
+
+		if opts.styling == nil {
+			// Try passing arbitrary data in, as anything should be accepted
+			opts.styling, _ = json.Marshal(map[string]any{
```
johnstcn, line 66: 👍 but we should also ensure that `nil` is accepted.

provider/formtype.go
```diff
@@ -0,0 +1,148 @@
...
+		OptionTypeBoolean,
+		OptionTypeListString,
+	}
+}
+
+// ParameterFormType is the list of supported form types for display in
+// the Coder "create workspace" form. These form types are functional as well
+// as cosmetic.
+// For example, "multi-select" has the type "list(string)" but the option
+// values are "string".
```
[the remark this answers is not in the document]
Emyrk: I'd rather defer that to the truth table below. I do not want to document it in 2 places.

provider/formtype.go
```diff
@@ -0,0 +1,148 @@
...
+	return paramType, specifiedFormType, nil
+}
+
+func toStrings[A ~string](l []A) []string {
+	var r []string
+	for _, v := range l {
+		r = append(r, string(v))
+	}
+	return r
+}
```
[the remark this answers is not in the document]
Emyrk: I debated it. We have it in the `slices` pkg of [`coder/coder`](https://github.com/coder/coder/blob/11bb88861cd81050f78bd67cca34f76eaa943947/coderd/util/slice/slice.go#L7-L14)

provider/formtype_test.go
Emyrk: If you do that, then the first failing `assert` stops the other assertions. 
I think I'd rather have each assertion dump it's failing output.

provider/formtype_test.go
Emyrk: If I rename it to `logDebugInfo`, then it's more accurate to what it is doing.

provider/parameter.go
Emyrk: I thought so too. I'll do it.

provider/formtype_test.go
Emyrk: Yup! That is the default: https://github.com/coder/terraform-provider-coder/blob/stevenmasley%2Fform_control/provider/parameter.go#L262-L262

johnstcn (approved):

PR #376, opened 2025-04-07
fix: correct the prebuilds type definition

SasSwart:
As documented in the updated code, prebuilds require a slice in the type definition to allow terraform parsing tools to use our defined types and avoid duplication or drift.


dannykopping (approved):
LGTM

PR #379, opened 2025-04-11
chore: enhance parameter validation error messages 

Emyrk:
Validation code is exported for external use

### DIFF
--- a/provider/formtype.go
+++ b/provider/formtype.go
@@ -115,9 +115,11 @@ var formTypeTruthTable = map[OptionType]map[bool][]ParameterFormType{
 // The use case is when using multi-select. The options are 'string' and the
 // value is 'list(string)'.
 func ValidateFormType(paramType OptionType, optionCount int, specifiedFormType ParameterFormType) (OptionType, ParameterFormType, error) {
-	allowed, ok := formTypeTruthTable[paramType][optionCount > 0]
+	optionsExist := optionCount > 0
+	allowed, ok := formTypeTruthTable[paramType][optionsExist]
 	if !ok || len(allowed) == 0 {
-		return paramType, specifiedFormType, xerrors.Errorf("value type %q is not supported for 'form_types'", paramType)
+		// This error should really never be hit, as the provider sdk does an enum validation.
+		return paramType, specifiedFormType, xerrors.Errorf("\"type\" attribute=%q is not supported, choose one of %v", paramType, OptionTypes())
 	}
 
 	if specifiedFormType == ParameterFormTypeDefault {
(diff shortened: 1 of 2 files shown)


provider/parameter.go
```diff
@@ -459,6 +412,93 @@ func valueIsType(typ OptionType, value string) diag.Diagnostics {
...
+		return singleDiag(diag.Diagnostic{
+			Severity:      diag.Error,
+			Summary:       "Invalid form_type for parameter",
+			Detail:        err.Error(),
+			AttributePath: cty.Path{cty.GetAttrStep{Name: "form_type"}},
+		})
+	}
+
+	optionNames := map[string]interface{}{}
+	optionValues := map[string]interface{}{}
```
aslilac, line 429:
```suggestion
	optionNames := map[string]any{}
	optionValues := map[string]any{}
```

aslilac (approved):

provider/formtype.go
```diff
@@ -126,7 +128,27 @@ func ValidateFormType(paramType OptionType, optionCount int, specifiedFormType P
...
+		// It serves as a more helpful error message.
+		//
+		// Eg: form_type=slider is valid for type=number, but invalid if options exist.
+		// And this error message is more accurate than just saying "form_type=slider is
+		// not valid for type=number".
+		if slices.Contains(opposite, specifiedFormType) {
+			if optionsExist {
+				optionMsg = " when options exist"
+			} else {
+				optionMsg = " when options do no exist"
```
johnstcn, line 145:
```suggestion
				optionMsg = " when options do not exist"
```

provider/parameter.go
```diff
@@ -459,6 +412,93 @@ func valueIsType(typ OptionType, value string) diag.Diagnostics {
...
+			if !defaultIsValid {
+				return singleDiag(diag.Diagnostic{
+					Severity:      diag.Error,
+					Summary:       "Default value must be a valid option",
+					Detail:        fmt.Sprintf("the value %q must be defined as one of options", v.Default),
+					AttributePath: cty.Path{cty.GetAttrStep{Name: "default"}},
+				})
+			}
+		}
+	}
```
johnstcn, line 497: suggestion: move this block up so we can exit early and then continue the rest of the validation knowing that `len(v.Option) > 0`.

johnstcn (approved):

provider/parameter.go
Emyrk: The `if...else` blocks still exists. I can bring this `if...else` 1 scope up if I add an early exit on `default != "" || len(v.Option) == 0`. I don't think it is much of an improvement.

I am going to leave it as it was for now.

PR #381, opened 2025-04-14
feat: parameter validation refactored to single function

Emyrk:
# What this PR does

Centralizes all validation logic into a single function on the parameter. This is intended to be imported by other packages (eg: preview) to do clientside validation before running terraform.

# Validation changes

See diff from `main`. Left side is the tests passing on `main`.
https://www.diffchecker.com/GaUlaTze/

- Input values **must** be in the option set. [Before](https://github.com/coder/terraform-provider-coder/blob/a2bb4c074b3257ac7e2fd361b7dea1cadf954206/provider/testdata/parameter_table.md?plain=1#L49) this was allowed, but enforced in `coder/coder`
- Number parameters require values to be numbers. [Before](https://github.com/coder/terraform-provider-coder/blob/a2bb4c074b3257ac7e2fd361b7dea1cadf954206/provider/testdata/parameter_table.md?plain=1#L43) this was allowed, but enforced in `coder/coder`

# Future work 

- Disallow empty values for creating workspaces: https://github.com/coder/terraform-provider-coder/blob/247df4e410c3c9ca159d59a8428975e5e873dc67/provider/parameter.go#L421-L427
- coder/coder to set the correct validation mode
- validate `monotonic` in the provider, passing in the old value via env var

[empty]: https://github.com/coder/terraform-provider-coder/blob/a2bb4c074b3257ac7e2fd361b7dea1cadf954206/provider/testdata/parameter_table.md?plain=1#L4-L14
[option]: https://github.com/coder/terraform-provider-coder/blob/a2bb4c074b3257ac7e2fd361b7dea1cadf954206/provider/testdata/parameter_table.md?plain=1#L20
[default]: https://github.com/coder/terraform-provider-coder/blob/a2bb4c074b3257ac7e2fd361b7dea1cadf954206/provider/testdata/parameter_table.md?plain=1#L46

### DIFF
--- a/provider/parameter.go
+++ b/provider/parameter.go
@@ -21,22 +21,6 @@ import (
 	"golang.org/x/xerrors"
 )
 
-type ValidationMode string
-
-const (
-	ValidationModeEnvVar = "CODER_VALIDATION_MODE"
-	// ValidationModeDefault is used for creating a workspace. It validates the final
-	// value used for a parameter. Some allowances for invalid options are tolerated,
-	// as unused options do not affect the final parameter value. The default value
-	// is also ignored, assuming a value is provided.
-	ValidationModeDefault ValidationMode = ""
-	// ValidationModeTemplateImport tolerates empty values as long as no 'validation'
-	// block is specified. This allows for importing a template without having to
-	// specify a value for every parameter. It does not tolerate an invalid default
-	// or invalid option values.
-	ValidationModeTemplateImport ValidationMode = "template-import"
-)
-
 var (
 	defaultValuePath = cty.Path{cty.GetAttrStep{Name: "default"}}
 )
@@ -160,8 +144,7 @@ func parameterDataSource() *schema.Resource {
 				input = &envValue
 			}
 
-			mode := os.Getenv(ValidationModeEnvVar)
-			value, diags := parameter.ValidateInput(input, ValidationMode(mode))
+			value, diags := parameter.ValidateInput(input)
 			if diags.HasError() {
 				return diags
 			}
@@ -410,17 +393,7 @@ func valueIsType(typ OptionType, value string) error {
 	return nil
 }
 
-func (v *Parameter) ValidateInput(input *string, mode ValidationMode) (string, diag.Diagnostics) {
-	if mode != ValidationModeDefault && mode != ValidationModeTemplateImport {
-		return "", diag.Diagnostics{
-			{
-				Severity: diag.Error,
-				Summary:  "Invalid validation mode",
-				Detail:   fmt.Sprintf("validation mode %q is not supported, use %q, or %q", mode, ValidationModeDefault, ValidationModeTemplateImport),
-			},
-		}
-	}
-
+func (v *Parameter) ValidateInput(input *string) (string, diag.Diagnostics) {
 	var err error
 	var optionType OptionType
 
@@ -443,29 +416,9 @@ func (v *Parameter) ValidateInput(input *string, mode ValidationMode) (string, d
 		}
 	}
 
-	optionValues, diags := v.ValidOptions(optionType, mode)
+	optionValues, diags := v.ValidOptions(optionType)
 	if diags.HasError() {
 		return "", diags
-	}
-
-	if mode == ValidationModeTemplateImport && v.Default != nil {
-		// Template import should validate the default value.
-		err := valueIsType(v.Type, *v.Default)
-		if err != nil {
-			return "", diag.Diagnostics{
-				{
-					Severity:      diag.Error,
-					Summary:       fmt.Sprintf("Default value is not of type %q", v.Type),
-					Detail:        err.Error(),
-					AttributePath: defaultValuePath,
-				},
-			}
-		}
-
-		d := v.validValue(*v.Default, optionType, optionValues, defaultValuePath)
-		if d.HasError() {
-			return "", d
-		}
 	}
 
 	// TODO: This is a bit of a hack. The current behavior states if validation
@@ -502,7 +455,7 @@ func (v *Parameter) ValidateInput(input *string, mode ValidationMode) (string, d
 	return forcedValue, nil
 }
 
-func (v *Parameter) ValidOptions(optionType OptionType, mode ValidationMode) (map[string]struct{}, diag.Diagnostics) {
+func (v *Parameter) ValidOptions(optionType OptionType) (map[string]struct{}, diag.Diagnostics) {
 	optionNames := map[string]struct{}{}
 	optionValues := map[string]struct{}{}
 
@@ -538,18 +491,10 @@ func (v *Parameter) ValidOptions(optionType OptionType, mode ValidationMode) (ma
 		optionValues[option.Value] = struct{}{}
 		optionNames[option.Name] = struct{}{}
 
-		if mode == ValidationModeTemplateImport {
-			opDiags := v.validValue(option.Value, optionType, nil, cty.Path{})
-			if opDiags.HasError() {
-				for i := range opDiags {
-					opDiags[i].Summary = fmt.Sprintf("Option %q: %s", option.Name, opDiags[i].Summary)
-				}
-				diags = append(diags, opDiags...)
-			}
-		}
+		// Option values are assumed to be valid. Do not call validValue on them.
 	}
 
-	if diags.HasError() {
+	if diags != nil && diags.HasError() {
 		return nil, diags
 	}
 	return optionValues, nil
--- a/provider/parameter.go
+++ b/provider/parameter.go
@@ -30,9 +30,10 @@ const (
 	// as unused options do not affect the final parameter value. The default value
 	// is also ignored, assuming a value is provided.
 	ValidationModeDefault ValidationMode = ""
-	// ValidationModeTemplateImport tolerates empty values, as the value might not be
-	// available at import. It does not tolerate an invalid default or invalid option
-	// values.
+	// ValidationModeTemplateImport tolerates empty values as long as no 'validation'
+	// block is specified. This allows for importing a template without having to
+	// specify a value for every parameter. It does not tolerate an invalid default
+	// or invalid option values.
 	ValidationModeTemplateImport ValidationMode = "template-import"
 )
 
@@ -428,14 +429,6 @@ func (v *Parameter) Valid(input *string, mode ValidationMode) (string, diag.Diag
 		value = v.Default
 	}
 
-	// TODO: When empty values want to be rejected, uncomment this.
-	//   coder/coder should update to use template import mode first,
-	//   before this is uncommented.
-	//if value == nil && mode == ValidationModeDefault {
-	//	var empty string
-	//	value = &empty
-	//}
-
 	// optionType might differ from parameter.Type. This is ok, and parameter.Type
 	// should be used for the value type, and optionType for options.
 	optionType, v.FormType, err = ValidateFormType(v.Type, len(v.Option), v.FormType)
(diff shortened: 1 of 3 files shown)


Emyrk:
@johnstcn @matifali should this pass a template import?

Before this PR, this would pass a `terraform plan`.

```hcl
data "coder_parameter" "region" {
  name        = "region"
  description = "Which region would you like to deploy to?"
  type        = "string"
  order       = 1

  option {
    name  = "Europe"
    value = "eu"
  }
  option {
    name  = "United States"
    value = "us"
  }
}
```

This PR sets the error to:

```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Value must be a valid option. The value is empty, did you forget to set it with a default or from user input?
│ 
│   with data.coder_parameter.region,
│   on main.tf line 21, in data "coder_parameter" "region":
│   21: data "coder_parameter" "region" {
│ 
│ the value "" must be defined as one of options
```

To fix this, either the value must be passed via an env var:

```
CODER_PARAMETER_c697d2981bf416569a16cfbcdec1542b5398f3cc77d2b905819aa99c46ecf6f6=us terraform plan
```

Or with a default:

```
data "coder_parameter" "region" {
  # ...
  default = "us"
  # ...
}
```

--------

I feel like we should reject it? But that is a breaking change.

matifali:
There are use cases where an admin may not want to set a default to force the user to choose an option.

johnstcn:
> There are use cases where an admin may not want to set a default to force the user to choose an option. 

This behaviour is problematic because you could end up with a plan that does not correspond to the apply, depending on how the value of the parameter is used.

Would marking the parameter as `ephemeral` be a possible workaround here?

matifali:
Do we require a user input if the parameters are ephemeral? Do we not allow setting a default value for such parameters?
Do such ephemeral parameters support options too?

Emyrk:
@matifali @johnstcn 

How I see it is the relaxed validation on template import is incorrect. If a parameter has 0 values, but will have some value at workspace create, then the `terraform plan` could be incorrect.

However, I understand the use case. The provider SDK we are using does not indicate anywhere if `terraform plan` vs `terraform apply` is being run. So there is no way to conditionally enforce validation that way.

My gut thought is that we need to effectively have 2 modes of validation. 1 for template import, 1 for workspace create, where the latter is strict. The strict validation is the most correct.

What if we pass through an env var a relaxed validation mode? `coder/coder` can pass `CODER_VALIDATION=relaxed` when doing a template import. Then the provider can accept `null` values.

spikecurtis:
> As for folks setting this env var at home, I think that would just be on them. By default the validation should be handled correctly in coder (when to be relaxed, and when not)

If provisionerd _always_ sets the environment variable before exec'ing terraform, then I don't think we need to worry about users setting it.

provider/parameter.go
```diff
@@ -429,84 +450,201 @@ func (v *Parameter) Valid() diag.Diagnostics {
...
+						Summary:       fmt.Sprintf("%s must be a valid option%s", name, extra),
+						Detail:        fmt.Sprintf("the value %q must be defined as one of options", value),
+						AttributePath: path,
 					},
 				}
 			}
 		}
 	}
 
+	if len(v.Validation) == 1 {
```
spikecurtis, line 633: Is there somewhere we enforce that there can only be a max of one Validation?

provider/parameter.go
```diff
@@ -429,84 +450,201 @@ func (v *Parameter) Valid() diag.Diagnostics {
...
+	}
+
+	// TODO: This is a bit of a hack. The current behavior states if validation
+	//   is given, then apply validation to unset values.
+	//   This should be removed, and all values should be validated. Meaning
+	//   value == nil should not be accepted in the first place.
+	if len(v.Validation) > 0 && value == nil {
+		empty := ""
+		value = &empty
+	}
```
spikecurtis, line 485: I feel like this would be more readable if you inverted the `if` clause and had it `return` in the case that we don't do validation.

provider/parameter.go
```diff
@@ -429,84 +450,201 @@ func (v *Parameter) Valid() diag.Diagnostics {
...
-		for _, option := range v.Option {
-			_, exists := optionNames[option.Name]
-			if exists {
-				return diag.Diagnostics{{
+	optionValues, diags := v.ValidOptions(optionType, mode)
+	if diags.HasError() {
+		return "", diags
+	}
+
+	if mode == ValidationModeTemplateImport && v.Default != nil {
```
spikecurtis, line 458: This is a weird case where template import is _more_ strict than default. What is the justification for that? I thought we were only making template import less strict so that we don't break people.

provider/parameter.go
```diff
@@ -21,6 +21,25 @@ import (
...
+
+const (
+	ValidationModeEnvVar = "CODER_VALIDATION_MODE"
+	// ValidationModeDefault is used for creating a workspace. It validates the final
+	// value used for a parameter. Some allowances for invalid options are tolerated,
+	// as unused options do not affect the final parameter value. The default value
+	// is also ignored, assuming a value is provided.
+	ValidationModeDefault ValidationMode = ""
+	// ValidationModeTemplateImport tolerates empty values, as the value might not be
+	// available at import. It does not tolerate an invalid default or invalid option
```
spikecurtis, line 34: IIUC, It doesn't tolerate empty values if there is a validation defined, so this is misleading.

provider/parameter.go
```diff
@@ -144,38 +153,24 @@ func parameterDataSource() *schema.Resource {
...
+				input = &envValue
 			}
-			// Set the form_type back in case the value was changed.
-			// Eg via a default. If a user does not specify, a default value
-			// is used and saved.
-			rd.Set("form_type", parameter.FormType)
 
-			diags := parameter.Valid()
+			mode := os.Getenv(ValidationModeEnvVar)
+			value, diags := parameter.Valid(input, ValidationMode(mode))
```
spikecurtis, line 163: Confusing name. Conceptually there is a Parameter resource definition, which might be invalid, and a parameter value, which might also be invalid. This is the latter, but it's defined on the struct that represents the former, so it's confusing. Maybe `parameter.ValidateInput`?

provider/parameter.go
```diff
@@ -389,37 +384,63 @@ func fixValidationResourceData(rawConfig cty.Value, validation interface{}) (int
...
+		value = v.Default
+	}
+
+	// TODO: When empty values want to be rejected, uncomment this.
+	//   coder/coder should update to use template import mode first,
+	//   before this is uncommented.
+	//if value == nil && mode == ValidationModeDefault {
+	//	var empty string
+	//	value = &empty
+	//}
```
spikecurtis, line 437: This business wouldn't be necessary if the provider assumes the "looser" validation mode if unset. That is, there are 2 modes: WorkspaceCreate and TemplateImport, which in future we will explicitly set via environment variable. If unset, we assume TemplateImport, which has no breaking changes from current behavior.

spikecurtis (commented):
It seems very odd to me that in this PR, template-import mode, which is supposedly being created for back-compatibility is _more_ strict than what you're calling default mode.

Shouldn't it be the other way around? That template-import mode skips some checks that would break backward compatibility?

Emyrk:
> It seems very odd to me that in this PR, template-import mode, which is supposedly being created for back-compatibility is more strict than what you're calling default mode.

> Shouldn't it be the other way around? That template-import mode skips some checks that would break backward compatibility?

Maybe I should have broken this into 2 PRs. The backwards compatibility I am keeping in tact is allowing empty an `default` value. So this is **still** valid:

```terraform
data "coder_parameter" "region" {
  name        = "region"
  description = "Which region would you like to deploy to?"
  type        = "string"
  order       = 1

  option {
    name  = "Europe"
    value = "eu"
  }
}
```

---

The changes I am making that are more strict are the checks in `coder/coder`. So yes, this change shifts some of the validation in `coder/coder` -> `terraform-provider-coder`.

This is now invalid except at `template-import`, since no input values are given. So at `template-import`, there is no change in behavior.
```terraform
# Using input value 3
# CODER_PARAMETER_f3c7807d475073ba009bf4801b2d934e9f0126cb96dd19a27dbffcae23a7f5a3=3 terraform plan

data "coder_parameter" "numbers" {
  name        = "numebrs"
  type        = "number"
  order       = 1

  option {
    name  = "Five"
    value = "5"
  }
}
```

provider/parameter.go
Emyrk: I can do that :+1:

provider/parameter.go
Emyrk: You are correct. When we turn this on in coderd, **new** templates imported will be subject to a stricter validation. There is 2 reasons.

## Strict options

You can currently add `option`s to your parameters that are invalid. That is a bit confusing on the UI, when options exist, but if you select them, it would fail to build.

At `create` time, the options are ignored, since they are not used.

## Strict Default

This was actually done before too **iff** no input value was given (which is usually the case for template-import). I made this more explicit, where the `Default` has to be valid if set. Even if you passed in an input value at import time.

I enforce this, because if you do not, then a user who leaves a default value for a form has an invalid value and cannot build.

----

Sidenote, but if you want to place a value in the UI without it being valid, we are adding:

```terraform
data "coder_parameter" "numbers" {
  name        = "numebrs"
  type        = "number"
  order       = 1

  styling = jsonencode({
    placeholder = "Enter a number"
  })
}
```

provider/parameter.go
Emyrk: Correct. I will update

provider/parameter.go
Emyrk: I removed this and did the early `return` instead as you suggested.

`TemplateImport` actually has more breaking changes than `create` in the end. `create` is backwards compatible because imported templates already exist and must still work.

Newly imported templates (once coderd switches), is going to require every `option` to also be valid. By doing this, the workspace create form will only present valid options to the user.


Eventually I would like to disallow an empty input value at `create` entirely.

Emyrk:
I removed the idea of 2 validation modes. There is only 1. `option` values are not validated, just as they were previously skipped.

The new validation mode is backwards compatible in all intended behavior.

provider/parameter.go
```diff
@@ -144,38 +138,23 @@ func parameterDataSource() *schema.Resource {
...
-			diags := parameter.Valid()
+			value, diags := parameter.ValidateInput(input)
 			if diags.HasError() {
 				return diags
 			}
 
+			// Always set back the value, as it can be sourced from the default
+			rd.Set("value", value)
+
+			// Set the form_type as it could have changed in the validation.
```
spikecurtis, line 155: I found this curious and confusing. A fuller explanation is probably worth setting out here: the `form_type` can start off unset, in which case the "validation" code selects a form type based on the `type`.

It's also a bit of a naming failure that we call the procedure that does this "Validation", which implies it doesn't change anything and just renders a verdict. In the Coderd code, I believe we called this procedure "resolution".

provider/parameter.go
```diff
@@ -429,84 +416,170 @@ func (v *Parameter) Valid() diag.Diagnostics {
...
+
+	// First validate if the value is a valid option
+	if len(optionValues) > 0 {
 		if v.Type == OptionTypeListString && optionType == OptionTypeString {
 			// If the type is list(string) and optionType is string, we have
 			// to ensure all elements of the default exist as options.
-			defaultValues, diags := valueIsListString(v.Default, cty.Path{cty.GetAttrStep{Name: "default"}})
-			if diags.HasError() {
-				return diags
+			listValues, err := valueIsListString(value)
```
spikecurtis, line 515: I think the comment above is out of date, as we are not specifically checking the default anymore

provider/parameter.go
```diff
@@ -429,84 +416,170 @@ func (v *Parameter) Valid() diag.Diagnostics {
...
+		// Option values are assumed to be valid. Do not call validValue on them.
+	}
+
+	if diags != nil && diags.HasError() {
+		return nil, diags
+	}
+	return optionValues, nil
+}
+
+func (v *Parameter) validValue(value string, optionType OptionType, optionValues map[string]struct{}, path cty.Path) diag.Diagnostics {
```
spikecurtis, line 503: this function is called only once, with an empty path.

provider/parameter.go
```diff
@@ -429,84 +416,170 @@ func (v *Parameter) Valid() diag.Diagnostics {
...
-			if diags.HasError() {
-				return diags
+			listValues, err := valueIsListString(value)
+			if err != nil {
+				return diag.Diagnostics{
+					{
+						Severity:      diag.Error,
+						Summary:       "When using list(string) type, value must be a json encoded list of strings",
+						Detail:        err.Error(),
+						AttributePath: defaultValuePath,
```
spikecurtis, line 522: We don't specifically know this is the default; it could just be the normal input.

provider/parameter.go
```diff
@@ -429,84 +416,170 @@ func (v *Parameter) Valid() diag.Diagnostics {
...
-						Summary:  "Default values must be a valid option",
+						Summary:  fmt.Sprintf("%ss must be a valid option", name),
 						Detail: fmt.Sprintf(
-							"default value %q is not a valid option, values %q are missing from the options",
-							v.Default, strings.Join(missing, ", "),
+							"%s %q is not a valid option, values %q are missing from the options",
+							name, value, strings.Join(missing, ", "),
 						),
-						AttributePath: cty.Path{cty.GetAttrStep{Name: "default"}},
+						AttributePath: defaultValuePath,
```
spikecurtis, line 545: Again, this might not be the default.

spikecurtis (approved):
some comments and suggestions inline, but I don't need to review again.

provider/parameter.go
Emyrk: It is unfortunate. I updated the comment.

provider/parameter.go
Emyrk: Updated the call to pass in the default value path if the value is sourced from the default.

The path (if it can be specified), makes the error a bit nicer:

Screenshot From 2025-05-07 07-50-11


Without path

Screenshot From 2025-05-07 07-50-16

PR #387, opened 2025-04-30
test: unit test to document validation behavior of parameters

Emyrk:
Locking in the behavior of validations before I change things. My goal is to move all validation logic into this provider. At present, some of it is done in `coder/coder`.

If you see https://github.com/coder/terraform-provider-coder/pull/381, adding more strict validation would be a regression in behavior. So I might have to add 2 validation modes so that template import can still be done with relaxed valiation.

Regardless, this test will help illustrate what I change when I add stricter validation.

I confirmed this test passes on a commit 3 weeks ago, before any of my refactors.

Some things we should probably fix or document:
- The default value can be invalid if an input value is valid. So the default is not checked.
- The value input does not need to be in the optionset, this is enforced in `coder/coder`

**NOTE:** Monotonic is not yet checked, it should be added into the terraform


provider/parameter_test.go
```diff
@@ -686,6 +687,259 @@ data "coder_parameter" "region" {
...
+func TestParameterValidationEnforcement(t *testing.T) {
+	// Some interesting observations:
+	// - Validation logic does not apply to the value of 'options'
+	//	- [NumDefInvOpt] So an invalid option can be present and selected, but would fail
+	// - Validation logic does not apply to the default if a value is given
+	//	- [NumIns/DefInv] So the default can be invalid if an input value is valid.
+	//	  The value is therefore not really optional, but it is marked as such.
+	// - [NumInsNotOptsVal | NumsInsNotOpts] values do not need to be in the option set?
+
+	table := strings.TrimSpace(`
```
johnstcn, line 707: This table is a bit over the upper limit of what I'd consider "comfortable" to read inline. Might be better extracted to its own file. Bonus: it becomes previewable in GH!

provider/parameter_test.go
```diff
@@ -686,6 +687,259 @@ data "coder_parameter" "region" {
...
+| Name                | Type          | Input Value | Default | Options           | Validation | -> | Output Value | Optional | Error        |
+|---------------------|---------------|-------------|---------|-------------------|------------|----|--------------|----------|--------------|
+|                     | Empty Vals    |             |         |                   |            |    |              |          |              |
+| Emty                | string,number |             |         |                   |            |    | ""           | false    |              |
```
johnstcn, line 711: typo: Empty

provider/parameter_test.go
```diff
@@ -686,6 +687,259 @@ data "coder_parameter" "region" {
...
+					Min:         0,
+					MinDisabled: true,
+					Max:         0,
+					MaxDisabled: true,
+					Monotonic:   "",
+					Regex:       columns[5],
+					Error:       "regex error",
+				}
+			}
+		}
```
johnstcn, line 831: suggestion: Extract this to a `testutil` package maybe?

johnstcn (approved):
Some comments below, but these are optional for a new unit test.

PR #392, opened 2025-05-06
feat: enforce monotonicity in terraform provider 

Emyrk:
Previous value must come from env var. This change has to be made in `coder/coder` to take advantage of this.

[`CODER_PARAMETER_PREVIOUS_<hash>`](https://github.com/coder/terraform-provider-coder/blob/e011f733bf27b0f3daf58838fc3c4531d7b4e36c/provider/parameter.go#L697-L697)

Unfortunately reading from `tfstate` requires the block to be a `resource` block. This is because a `data` block only has access to the `Read` context.

https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource.go#L287-L354
Read
> // The *ResourceData parameter contains the state data for this managed
// resource instance or data resource instance.

Update
> // The *ResourceData parameter contains the plan and state data for this
// managed resource instance. The available data in the Get* methods is the
// the proposed state, which is the merged data of the prior state,
// practitioner configuration, and any CustomizeDiff field logic. The
// available data for the GetChange* and HasChange* methods is the prior
// state and proposed state.

`coder/coder` PR in progress https://github.com/coder/coder/pull/17696

### DIFF
--- a/provider/testdata/parameter_table.md
+++ b/provider/testdata/parameter_table.md
@@ -28,6 +28,8 @@
 | NumDefDecBad         | number        | 4        |           | 5       |                   | decreasing |    |        |          | lower           |
 | NumDefDecEq          | number        | 5        |           | 5       |                   | decreasing |    | 5      | true     |                 |
 | NumDefIncEq          | number        | 5        |           | 5       |                   | increasing |    | 5      | true     |                 |
+| NumDefIncNaN         | number        | a        |           | 5       |                   | increasing |    | 5      | true     |                 |
+| NumDefDecNaN         | number        | b        |           | 5       |                   | decreasing |    | 5      | true     |                 |
 |                      |               |          |           |         |                   |            |    |        |          |                 |
 | StrDef               | string        |          |           | hello   |                   |            |    | hello  | true     |                 |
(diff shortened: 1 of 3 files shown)


provider/testdata/parameter_table.md
```diff
@@ -1,80 +1,93 @@
...
+| NumDefOptsNotNum     | number        |          |           | 1       | 1,a,2             |            |    |        |          | type "number"   |
+| NumDefInc            | number        | 4        |           | 5       |                   | increasing |    | 5      | true     |                 |
+| NumDefIncBad         | number        | 6        |           | 5       |                   | increasing |    |        |          | greater         |
+| NumDefDec            | number        | 6        |           | 5       |                   | decreasing |    | 5      | true     |                 |
+| NumDefDecBad         | number        | 4        |           | 5       |                   | decreasing |    |        |          | lower           |
+| NumDefDecEq          | number        | 5        |           | 5       |                   | decreasing |    | 5      | true     |                 |
+| NumDefIncEq          | number        | 5        |           | 5       |                   | increasing |    | 5      | true     |                 |
```
spikecurtis, line 30: test case for previous not a number?

spikecurtis (approved):

PR #398, opened 2025-05-12
chore: add tag_sort option to goreleaser

dannykopping:
Choose more recently-created tag if multiple tags point to same ref

I published `v2.4.1` pointing to the same ref as `v2.4.1-pre0` after validating the latter. However when I published `v2.4.1`, the [action](https://github.com/coder/terraform-provider-coder/actions/runs/14965521939) chose the older tag.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -60,3 +60,5 @@ changelog:
   use: github-native
   sort: asc
   abbrev: 0
+git:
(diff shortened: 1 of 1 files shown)


dannykopping:
Hhmm nope that didn't fix it https://github.com/coder/terraform-provider-coder/actions/runs/14965932298/job/42035952537

PR #421, opened 2025-07-21
fix: correct URL validation and centralize logic

ssncferreira:
### Description

Fixes incorrect URL validation logic previously used in URL definitions, where the attribute label (`s`) was being parsed instead of the actual value. This bug caused invalid URLs to be silently accepted.

Additionally, this PR introduces a shared `ValidateURL` helper function to remove duplication and ensure consistent behavior across all URL fields.

### Changes

- Fixed incorrect use of `url.Parse(s)`, now validating the actual value.
- Added `ValidateURL` helper function for consistent and reusable validation.
- Updated all schema definitions that perform URL validation to use `ValidateURL`.
- Adjusted tests to match the corrected validation logic.

### DIFF
--- a/provider/helpers/validation.go
+++ b/provider/helpers/validation.go
@@ -5,15 +5,18 @@ import (
 	"net/url"
 )
 
-// ValidateURL checks that the given value is a valid URL string.
+// ValidateURL validates that value is a valid URL string.
+// Accepts empty strings, local file paths, file:// URLs, and http/https URLs.
 // Example: for `icon = "/icon/region.svg"`, value is `/icon/region.svg` and label is `icon`.
-func ValidateURL(value interface{}, label string) ([]string, []error) {
+func ValidateURL(value any, label string) ([]string, []error) {
 	val, ok := value.(string)
 	if !ok {
 		return nil, []error{fmt.Errorf("expected %q to be a string", label)}
 	}
+
 	if _, err := url.Parse(val); err != nil {
 		return nil, []error{err}
 	}
+	
 	return nil, nil
 }
--- a/provider/helpers/validation.go
+++ b/provider/helpers/validation.go
@@ -1,0 +1,19 @@
+package helpers
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// ValidateURL checks that the given value is a valid URL string.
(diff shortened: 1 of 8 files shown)


provider/parameter.go
```diff
@@ -6,7 +6,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"net/url"
+	"github.com/coder/terraform-provider-coder/v2/provider/helpers"
```
johnstcn, line 9: nit: ordering

provider/helpers/validation.go
```diff
@@ -0,0 +1,19 @@
+package helpers
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// ValidateURL checks that the given value is a valid URL string.
+// Example: for `icon = "/icon/region.svg"`, value is `/icon/region.svg` and label is `icon`.
+func ValidateURL(value interface{}, label string) ([]string, []error) {
```
johnstcn, line 10: nti: `any` maybe?

provider/parameter_test.go
```diff
@@ -227,23 +227,19 @@ func TestParameter(t *testing.T) {
...
 				option {
 					name = "1"
 					value = "1"
 					icon = "/icon/code.svg"
 					description = "Something!"
 				}
-				option {
-					name = "2"
-					value = "2"
-				}
```
johnstcn, line 240: Is this change related?

provider/provider.go
```diff
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"github.com/coder/terraform-provider-coder/v2/provider/helpers"
```
johnstcn, line 5: nit: ordering

provider/parameter.go
ssncferreira: Addressed in https://github.com/coder/terraform-provider-coder/pull/421/commits/de3cf097faa71bf1512cfb325e9d4dd39d0999a2

provider/provider.go
ssncferreira: Addressed in https://github.com/coder/terraform-provider-coder/pull/421/commits/de3cf097faa71bf1512cfb325e9d4dd39d0999a2

johnstcn (approved):

PR #424, opened 2025-08-05
feat(coder-attach): add coder_external_agent resource

kacpersaw:
This PR introduces a new Terraform resource `coder_external_agent` that allows users to define external agents for use in workspaces. It allows the coder UI and coder/coder to detect and act on external agent presence.

This is part of https://github.com/coder/coder/issues/19091

### DIFF
--- a/provider/external_agent.go
+++ b/provider/external_agent.go
@@ -20,12 +20,11 @@ func externalAgentResource() *schema.Resource {
 		ReadContext:   schema.NoopContext,
 		DeleteContext: schema.NoopContext,
 		Schema: map[string]*schema.Schema{
-			"token": {
+			"agent_id": {
+				Type:        schema.TypeString,
+				Description: "The `id` property of a `coder_agent` resource to associate with.",
 				ForceNew:    true,
 				Required:    true,
-				Sensitive:   true,
-				Description: "Set the environment variable `CODER_AGENT_TOKEN` with this token to authenticate an agent.",
-				Type:        schema.TypeString,
 			},
 		},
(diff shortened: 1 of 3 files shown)


provider/external_agent_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExternalAgent(t *testing.T) {
+	t.Parallel()
+
+	t.Run("OK", func(t *testing.T) {
```
deansheather, line 14: Should also have a test that links this resource to a `coder_agent`.

provider/external_agent.go
```diff
@@ -0,0 +1,32 @@
...
+		ReadContext:   schema.NoopContext,
+		DeleteContext: schema.NoopContext,
+		Schema: map[string]*schema.Schema{
+			"token": {
+				ForceNew:    true,
+				Required:    true,
+				Sensitive:   true,
+				Description: "Set the environment variable `CODER_AGENT_TOKEN` with this token to authenticate an agent.",
+				Type:        schema.TypeString,
+			},
```
deansheather, line 29: I don't understand linking these based on the token value rather than just the ID value of the agent. It seems like we'd want to be able to configure the auth type on the `coder_agent` and have that also apply here.

deansheather (approved):

PR #435, opened 2025-08-29
feat: add tooltip field to workspace app

rafrdz:
## Summary

In this pull request we're adding an optional `tooltip` field. The `tooltip` field is a string field (with markdown support) that will be used to display tooltips on hover over app buttons in a workspace dashboard.

Tooltip screenshot

Screenshot 2025-08-29 at 4 11 56 PM

Tooltip video

https://github.com/user-attachments/assets/58c5d181-06bf-409f-b44c-5e92665664b4

Issue: https://github.com/coder/coder/issues/18431
Related coder PR: https://github.com/coder/coder/pull/19651

### Changes

- Added `tooltip` to the `app` schema with a max length of 2048 characters
- Updated `app` documentation to include the `tooltip` field
- Added an example `tooltip`

### Testing

- Added tests for empty, valid, and max length (to test validation logic)

### DIFF
--- a/docs/resources/app.md
+++ b/docs/resources/app.md
@@ -31,6 +31,7 @@ resource "coder_app" "code-server" {
   display_name = "VS Code"
   icon         = "${data.coder_workspace.me.access_url}/icon/code.svg"
   url          = "http://localhost:13337"
+  tooltip      = "You need to [Install Coder Desktop](https://coder.com/docs/user-guides/desktop#install-coder-desktop) to use this button."
   share        = "owner"
   subdomain    = false
   open_in      = "window"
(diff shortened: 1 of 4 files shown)


matifali:
@rafrdz This is great. Thank you.

matifali:
@rafrdz, one last thing to consider here is if someone used the latest provider with an older version of Coder, where this is not supported? Can we add a warning (or better, fail) in the provider to inform the user of the minimum required Coder version?

This warning/error should only happen if someone tries to use `tooltip` on an older Coder version. 
 
If this looks too much work, let me know and we can handle this gracefully as a noop in older Coder versions (better than nothing)

PR #438, opened 2025-09-05
ci: fix go version to that defined in go.mod

johnstcn:
Ref: https://github.com/coder/terraform-provider-coder/pull/437

Our version of Go we reference in our `.github/**` doesn't match that in `go.mod`.
To fix this, I've updated the version of Go and added a check to ensure they're all in sync.

Disclaimer: Most of the script was generated by ChatGPT but I went over it before committing.

### DIFF
--- a/scripts/check_go_version.sh
+++ b/scripts/check_go_version.sh
@@ -1,34 +1,15 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-fix=0
-args="$(getopt -o "" -l fix -- "$@")"
-eval set -- "$args"
-while true; do
-	case "$1" in
-	--fix)
-		fix=1
-        shift
-		;;
-	--)
-		shift
-		break
-		;;
-	*)
-		error "Unrecognized option: $1"
-		;;
-	esac
-done
-
 MOD_VERSION=$(go mod edit -json | jq -r .Go)
 echo "go.mod version: $MOD_VERSION"
 STATUS=0
 
-if [[ $fix -eq 1 ]]; then
+if [[ " $* " == *" --fix "* ]]; then
     for wf in .github/workflows/*.{yml,yaml}; do
         sed -i "s/go-version:.*/go-version: \"${MOD_VERSION}\"/g" "${wf}"
     done
-    exit $STATUS
+    exit 0
 fi
 
 for wf in .github/workflows/*.{yml,yaml}; do
@@ -37,9 +18,9 @@ for wf in .github/workflows/*.{yml,yaml}; do
         continue
     fi
 
-    UNIQUE_WF_VERSIONS=$(echo "$WF_VERSIONS" | sort -u)
+    UNIQUE_WF_VERSIONS=$(sort -u <<<"$WF_VERSIONS")
     for ver in $UNIQUE_WF_VERSIONS; do
-        if [[ "${ver}" != "$MOD_VERSION" ]]; then
+        if [[ $ver != "$MOD_VERSION" ]]; then
             STATUS=1
             echo "❌ $wf: go.mod=$MOD_VERSION but workflow uses $(tr '\n' ' ' <<<"$UNIQUE_WF_VERSIONS")"
             continue


.github/workflows/release.yml
```diff
@@ -19,15 +19,18 @@ jobs:
...
+        uses: actions/checkout@v5
+
+      - name: Check Go Versions
+        run: ./scripts/check_go_version.sh
+
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.22"
+          go-version: "1.24.2"
```
mafredri, line 31: Why do these need this to be in sync? IIRC Go will run as the version defined in `go.mod` even if it's a lower version. Higher version of course takes precedence.

See:

```
~
❯ go1.23.7 version
go version go1.23.7 darwin/arm64
```

```
~/Code/coder main*
❯ go1.23.7 version
go version go1.24.6 darwin/arm64
```

Btw, `go.mod` says `1.24.6`.

scripts/check_go_version.sh
```diff
@@ -0,0 +1,32 @@
...
+set -euo pipefail
+
+MOD_VERSION=$(go mod edit -json | jq -r .Go)
+echo "go.mod version: $MOD_VERSION"
+
+for wf in .github/workflows/*.yml; do
+    echo "Checking $wf ..."
+
+    WF_VERSIONS=$(yq -r '.jobs[].steps[] | select(.with["go-version"]) | .with["go-version"]' -o=tsv "$wf" | grep -v '^---$' || true)
+    if [ -z "$WF_VERSIONS" ]; then
```
mafredri, line 11:
```suggestion
    if [[ -z "$WF_VERSIONS" ]]; then
```
We're using Bash, not sh 😄

scripts/check_go_version.sh
```diff
@@ -0,0 +1,32 @@
...
+        echo "ℹ️ No go-version found in $wf (skipped)"
+        continue
+    fi
+
+    UNIQUE_WF_VERSIONS=$(echo "$WF_VERSIONS" | sort -u)
+    if [ "$(echo "$UNIQUE_WF_VERSIONS" | wc -l)" -ne 1 ]; then
+        echo "❌ Multiple Go versions found in $wf:"
+        echo "$UNIQUE_WF_VERSIONS"
+        exit 1
```
mafredri, line 20: We could gather up all errors from all workflows and print or auto-correct everything at the end? It'll be annoying to run this script, get one error, fix, get another, etc.

scripts/check_go_version.sh
```diff
@@ -0,0 +1,32 @@
...
+    fi
+
+    # At this point there's only one unique Go version
+    if [ "$UNIQUE_WF_VERSIONS" != "$MOD_VERSION" ]; then
+        echo "❌ Mismatch in $wf: go.mod=$MOD_VERSION but workflow uses $UNIQUE_WF_VERSIONS"
```
mafredri, line 25: With the above suggestions, don't think we need different cases for 1 vs many incorrect errors. Just print a list of bad files and the version(s) they contain, so they can be easily fixed. Or even better auto-apply fix.

scripts/check_go_version.sh
johnstcn: I'm hesitant about auto-correcting this given that it's meant to be run as a CI step. Maybe an optional argument `--autofix`?

.github/workflows/release.yml
mafredri: Ignore my comment about 1.24.6 btw, originally thought this was a PR to coder/coder repo 😄

scripts/check_go_version.sh
johnstcn: OK, the output is now terser:

```
# no issues
go.mod version: 1.24.2

# go.mod mismatch
go.mod version: 1.24.3
❌ Mismatch in .github/workflows/release.yml: go.mod=1.24.3 but workflow uses 1.24.2 
❌ Mismatch in .github/workflows/test.yml: go.mod=1.24.3 but workflow uses 1.24.2 

# multiple versions in workflow
go.mod version: 1.24.2
❌ Mismatch in .github/workflows/test.yml: go.mod=1.24.2 but workflow uses 1.24.2 1.24.3 
```

scripts/check_go_version.sh
johnstcn: Partially addressed in 759a03e, auto-fixing may be a separate change

scripts/check_go_version.sh
```diff
@@ -0,0 +1,32 @@
...
+    echo "Checking $wf ..."
+
+    WF_VERSIONS=$(yq -r '.jobs[].steps[] | select(.with["go-version"]) | .with["go-version"]' -o=tsv "$wf" | grep -v '^---$' || true)
+    if [ -z "$WF_VERSIONS" ]; then
+        echo "ℹ️ No go-version found in $wf (skipped)"
+        continue
+    fi
+
+    UNIQUE_WF_VERSIONS=$(echo "$WF_VERSIONS" | sort -u)
+    if [ "$(echo "$UNIQUE_WF_VERSIONS" | wc -l)" -ne 1 ]; then
```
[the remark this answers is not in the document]
johnstcn: 759a03e

scripts/check_go_version.sh
mafredri: Nice 👍🏻

scripts/check_go_version.sh
```diff
@@ -0,0 +1,54 @@
...
+		;;
+	--)
+		shift
+		break
+		;;
+	*)
+		error "Unrecognized option: $1"
+		;;
+	esac
+done
```
mafredri, line 21: Since this won't work on macOS, may I suggest a simple:

```bash
if [[ " $* " == *" --fix "* ]]; then
```

? 😄

scripts/check_go_version.sh
```diff
@@ -0,0 +1,54 @@
...
+
+MOD_VERSION=$(go mod edit -json | jq -r .Go)
+echo "go.mod version: $MOD_VERSION"
+STATUS=0
+
+if [[ $fix -eq 1 ]]; then
+    for wf in .github/workflows/*.{yml,yaml}; do
+        sed -i "s/go-version:.*/go-version: \"${MOD_VERSION}\"/g" "${wf}"
+    done
+    exit $STATUS
```
mafredri, line 31:
```suggestion
    exit 0
```
Suggestion: Keep it simple as long as possible.

scripts/check_go_version.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+for wf in .github/workflows/*.{yml,yaml}; do
+    WF_VERSIONS=$(yq -r '.jobs[].steps[] | select(.with["go-version"]) | .with["go-version"]' -o=tsv "$wf" | grep -v '^---$' || true)
+    if [[ -z "$WF_VERSIONS" ]]; then
+        continue
+    fi
+
+    UNIQUE_WF_VERSIONS=$(echo "$WF_VERSIONS" | sort -u)
```
mafredri, line 40:
```suggestion
    UNIQUE_WF_VERSIONS=$(sort -u <<<"$WF_VERSIONS")
```
Pet peeve: No need for pipe.

scripts/check_go_version.sh
```diff
@@ -0,0 +1,54 @@
...
+    for ver in $UNIQUE_WF_VERSIONS; do
+        if [[ "${ver}" != "$MOD_VERSION" ]]; then
```
mafredri, line 42:
```suggestion
        if [[ $ver != "$MOD_VERSION" ]]; then
```
Left side doesn't need quotes with `[[`.

mafredri (approved):
Some small suggestions/improvements but otherwise LGTM. Thanks for adding the `--fix` option!

scripts/check_go_version.sh
johnstcn: Is there a significance to the extra spaces?

scripts/check_go_version.sh
johnstcn: 1547929

scripts/check_go_version.sh
johnstcn: 1547929

scripts/check_go_version.sh
mafredri: Yup, it's a trick so it can match args exactly, _and_ in any placement (start, middle, end).

scripts/check_go_version.sh
johnstcn: 1547929

PR #445, opened 2025-09-26
feat: add `prompt` and `app_id` fields to `coder_ai_task`

DanielleMaywood:
Closes https://github.com/coder/internal/issues/977

Adds one required field `app_id`, one read-only computed field `prompt`,
as well as deprecates the `sidebar_app` field.

Also updates the `id` field to inherit the value from `CODER_TASK_ID`
env variable

### DIFF
--- a/provider/ai_task.go
+++ b/provider/ai_task.go
@@ -95,7 +95,7 @@ func aiTaskResource() *schema.Resource {
 			},
 			"app_id": {
 				Type:          schema.TypeString,
-				Description:   "The ID of the coder_app resource that provides the AI interface for this task.",
+				Description:   "The ID of the `coder_app` resource that provides the AI interface for this task.",
 				ForceNew:      true,
 				Optional:      true,
 				Computed:      true,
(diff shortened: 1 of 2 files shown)


provider/ai_task.go
```diff
@@ -21,13 +22,43 @@ type AITaskSidebarApp struct {
...
 		SchemaVersion: 1,
 
 		Description: "Use this resource to define Coder tasks.",
 		CreateContext: func(c context.Context, resourceData *schema.ResourceData, i any) diag.Diagnostics {
-			resourceData.SetId(uuid.NewString())
+			if idStr := os.Getenv("CODER_TASK_ID"); idStr != "" {
+				resourceData.SetId(idStr)
+			} else {
+				resourceData.SetId(uuid.NewString())
+			}
```
johnstcn, line 35: Would it make more sense to exit with an error here?

johnstcn: The integration tests will tell us :)
I'd expect that it won't have any effect unless the template defines a `coder_ai_task`. In that case, folks can lock their provider version.

johnstcn (approved):

PR #447, opened 2025-10-01
fix: make `ai_task.id` change backwards compatible

DanielleMaywood:


provider/ai_task.go
```diff
@@ -29,10 +28,17 @@ func aiTaskResource() *schema.Resource {
...
+
 			if idStr := os.Getenv("CODER_TASK_ID"); idStr != "" {
 				resourceData.SetId(idStr)
 			} else {
-				return diag.Errorf("CODER_TASK_ID must be set")
+				resourceData.SetId(uuid.NewString())
+
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Warning,
+					Summary:  "`CODER_TASK_ID` should be set, this will be an error in the future",
```
johnstcn, line 40:
```suggestion
					Summary:  "`CODER_TASK_ID` should be set. If you are seeing this message, the version of the Coder Terraform provider you are using is likely too new for your current Coder version.",
```

DanielleMaywood:
So the integration test is kinda messy, but we have a chicken-and-the-egg problem of `coder/coder` not yet supporting the updated resource 😄

Issue #325, opened 2025-01-07
Add `order` to `coder_metadata`


stirby:
Agent metadata currently does not support ordering in the same way users can order [apps](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/app#order-7) or [agents](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/agent#order-1). We should use the existing syntax to allow ordering metadata items on any resource. Each resource should have an independent index.

### Agent metadata ordering
The existing implementation:

```terraform
resource "coder_agent" "dev" {
  os   = "linux"
  ...
  metadata {
    display_name = "CPU Usage"
    key          = "cpu_usage"
    script       = "coder stat cpu"
    interval     = 10
    timeout      = 1
    order        = 2
  }
  metadata {
    display_name = "RAM Usage"
    key          = "ram_usage"
    script       = "coder stat mem"
    interval     = 10
    timeout      = 1
    order        = 1
  }

  order = 1
}
``` 

### Resource Metadata ordering
Proposal:

```terraform
resource "coder_metadata" "pod_info" {
  count       = data.coder_workspace.me.start_count
  resource_id = kubernetes_pod.dev[0].id
  # (Enterprise-only) this resource consumes 200 quota units
  daily_cost = 200
  item {
    key   = "description"
    value = "This description will show up in the Coder dashboard."
  }
  item {
    key   = "pod_uid"
    value = kubernetes_pod.dev[0].uid


    # new!
    order = 1
  }
  item {
    key   = "public_key"
    value = tls_private_key.example_key_pair.public_key_openssh

    # new!
    order = 2
  }
}

PR #450, opened 2025-10-21
feat: add order to coder_metadata

BrunoQuaresma:
Closes https://github.com/coder/terraform-provider-coder/issues/325

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -100,11 +100,16 @@ func populateIsNull(resourceData *schema.ResourceData) (result interface{}, err 
 	var resultItems []interface{}
 	for _, item := range items {
 		key := valueAsString(item.GetAttr("key"))
+		order, err := valueAsInt(item.GetAttr("order"))
+		if err != nil {
+			return nil, xerrors.Errorf("unable to parse order for coder_metadata item %q: %w", key, err)
+		}
+
 		resultItem := map[string]interface{}{
 			"key":       key,
 			"value":     valueAsString(item.GetAttr("value")),
 			"sensitive": valueAsBool(item.GetAttr("sensitive")),
-			"order":     valueAsInt(item.GetAttr("order")),
+			"order":     order,
 		}
 		if item.GetAttr("value").IsNull() {
 			resultItem["is_null"] = true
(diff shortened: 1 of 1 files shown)


provider/metadata.go
```diff
@@ -103,6 +103,12 @@ func metadataResource() *schema.Resource {
 							ForceNew: true,
 							Computed: true,
 						},
+						"order": {
+							Type:        schema.TypeInt,
+							Description: "The order determines the position of item in the UI presentation. The lowest order is shown first and items with equal order are sorted by key (ascending order).",
+							ForceNew:    true,
+							Computed:    true,
```
johnstcn: If you set `Computed` on an attribute, it cannot be changed. Ref: https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#computed

johnstcn (changes requested):
You will need to update the tests and run `make fmt` / `make gen` in order for CI to pass.
My suggestion would also be to add/update tests in `./integration` for this change.

provider/metadata_test.go
```diff
@@ -30,6 +30,7 @@ func TestMetadata(t *testing.T) {
 					item {
 						key = "foo"
 						value = "bar"
+						order = 4
```
mtojek: Can you add more test cases?

1. All items have `order` present.
2. Order is negative.
3. Only a few items have `order` present

provider/metadata_test.go
```diff
@@ -66,6 +67,7 @@ func TestMetadata(t *testing.T) {
 					"item.0.key":       "foo",
 					"item.0.value":     "bar",
 					"item.0.sensitive": "false",
+					"item.0.order":     "4",
```
mtojek: BTW this is not enough for metadata items. See the comment on top of this [function](https://github.com/coder/terraform-provider-coder/blame/main/provider/provider.go#L102-L106).

PS. VS code debugger is very handy when dealing with Terraform schema.

BrunoQuaresma:
@johnstcn related to the integration tests, what `minVersion` should I set when testing new features?

johnstcn:
> @johnstcn related to the integration tests, what `minVersion` should I set when testing new features?

If you set `minVersion: "v0.0.0"` then the integration test will always run.
Otherwise, the test will only run based on the version of Coder against which the integration test is actually running.

In CI, the integration tests run against three different versions (from `go run ./scripts/coderversion.go`):
- `CODER_MAINLINE_VERSION`: latest mainline version
- `CODER_STABLE_VERSION`: latest stable version
- `CODER_OLDSTABLE_VERSION`: previous stable version

So basically if a particular integration test only makes sense to run from a particular Coder version onwards, you set `minVersion` to that version.

mtojek (commented):
Once tests pointed our by Cian are 👍 , feel free to proceed.

PR #458, opened 2025-10-31
fix: set default value for `coder_ai_task.prompt` to an empty string

DanielleMaywood:
Instead of defaulting to "default" for the prompt, we should just leave it empty.


mafredri (approved):
I suppose we don't have any requirement on this to be length > 0 which would break terraform plan/template build?

If not, empty is fine 👍🏻

PR #459, opened 2025-10-31
docs(coder_ai_task): add min coder version for `prompt` and `enabled`

DanielleMaywood:


provider/ai_task.go
```diff
@@ -109,7 +109,7 @@ func aiTaskResource() *schema.Resource {
 			},
 			"enabled": {
 				Type:        schema.TypeBool,
-				Description: "True when executing in a Coder Task context, false when in a Coder Workspace context",
+				Description: "True when executing in a Coder Task context, false when in a Coder Workspace context.\n\n-> The `enabled` field is only populated in Coder v2.28 and later.",
```
johnstcn:
```suggestion
				Description: "True when executing in a Coder Task context, false when in a Coder Workspace context.\n\n  -> The `enabled` field is only populated in Coder v2.28 and later.",
```

Issue #343, opened 2023-07-27
`coder_app` breaks when `command = "..."` and `subdomain = true`


ElioDiNino:
Version: [`v0.27.1+65583ec`](https://github.com/coder/coder/commit/65583eca47f6d9d194ed8b038c3c010a4cd331c4)

If you create a `coder_app` like the following, you get a 500 error from Coder.

Either support for subdomains with commands should be added or the Terraform provider should complain that `subdomain = true` isn't allowed when `command` is set. When setting the `subdomain` key to `false` (or removing it entirely) things work fine (apart from the URL not being a subdomain).

```terraform
resource "coder_app" "keeper-commander" {
  agent_id     = coder_agent.main.id
  slug         = "keeper-commander"
  display_name = "Keeper CLI"
  icon         = "https://www.keepersecurity.com/assets/branding/android-icon-192x192.png"
  command      = "keeper shell"
  subdomain    = true
  share        = "owner"
}
```

<img width="896" alt="Screenshot 2023-07-26 at 6 18 37 PM" src="https://github.com/coder/terraform-provider-coder/assets/52972357/4459b4f1-796e-446d-b0db-f9042e54560c">

PR #469, opened 2025-11-24
feat: add confliction with `subdomain`

jakehwll:
Addresses #343 

This pull-request ensures that when we have a `command` object key that it will conflict if we attempt to use `subdomain = true` specified. This was previously an error back in July 2023, however it seems that we no longer actually throw a 500 if this doesn't exist, rather we simply ignore the `subdomain` key. 

For better overall handling we simply conflict against this now.

### DIFF
--- a/provider/app.go
+++ b/provider/app.go
@@ -86,7 +86,7 @@ func appResource() *schema.Resource {
 				Description: "A command to run in a terminal opening this app. In the web, " +
 					"this will open in a new tab. In the CLI, this will SSH and execute the command. " +
 					"Either `command` or `url` may be specified, but not both. " +
-					"If `command` is specified, `subdomain` must be unset.",
+					"Conflicts with `subdomain`.",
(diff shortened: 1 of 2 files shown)


provider/app.go
```diff
@@ -85,8 +85,9 @@ func appResource() *schema.Resource {
 				Type: schema.TypeString,
 				Description: "A command to run in a terminal opening this app. In the web, " +
 					"this will open in a new tab. In the CLI, this will SSH and execute the command. " +
-					"Either `command` or `url` may be specified, but not both.",
-				ConflictsWith: []string{"url"},
+					"Either `command` or `url` may be specified, but not both. " +
+					"If `command` is specified, `subdomain` must be unset.",
+				ConflictsWith: []string{"url", "subdomain"},
```
mtojek: Actually, can you add a test case to `app_test.go` to exercise `ConflictsWith`?

PR #501, opened 2026-04-10
feat: add coder_secret data source for user secrets

zedkipp:
Add a new `coder_secret` data source that allows template authors to declare required user secrets and access their values during workspace builds.

Schema:
- `env` (optional) — environment variable name the secret injects
- `file` (optional) — file path the secret injects
- `help_message` (required) — guidance shown when the secret is missing
- `value` (computed, sensitive) — resolved from provisioner env vars

Exactly one of `env` or `file` must be set. On start transitions, a missing secret fails the build with the help_message. On stop/delete, missing secrets return empty to allow teardown.

Env var convention:
- Env secrets: CODER_SECRET_ENV_{env_name}
- File secrets: CODER_SECRET_FILE_{hex(file_path)}

RFC: https://www.notion.so/coderhq/User-Secrets-32cd579be59280a98192dec7b3b5daf1

Related coder PR: https://github.com/coder/coder/pull/24542


provider/provider.go
```diff
@@ -64,6 +64,7 @@ func New() *schema.Provider {
 			"coder_external_auth":    externalAuthDataSource(),
 			"coder_workspace_owner":  workspaceOwnerDataSource(),
 			"coder_workspace_preset": workspacePresetDataSource(),
+			"coder_secret":           secretDataSource(),
```
zedkipp: The order seems arbitrary here. Just append to bottom of the list.

examples/data-sources/coder_secret/data-source.tf
zedkipp, line 1: Drop the specific examples (e.g. github) and make arbitrary

provider/secret.go
```diff
@@ -0,0 +1,132 @@
...
+					_, _ = fmt.Fprintf(&detail, "Required: %s\n\n", requirement)
+					if helpMessage := rd.Get("help_message").(string); helpMessage != "" {
+						_, _ = fmt.Fprintf(&detail, "Help message: %s\n\n", helpMessage)
+					}
+					_, _ = fmt.Fprintf(&detail, "To resolve: ensure a secret exposes the %s.\n", requirement)
+
+					return diag.Diagnostics{{
+						Severity: diag.Error,
+						Summary:  fmt.Sprintf("Missing required secret: %s", requirement),
+						Detail:   detail.String(),
```
zedkipp, line 82: This is what the diagnostic looks like all together:

Image

provider/secret.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+			if env != "" && file != "" {
+				return diag.Errorf("exactly one of `env` or `file` must be set")
+			}
+
+			// Build a stable ID from whichever field is set.
+			if env != "" {
+				rd.SetId(fmt.Sprintf("env:%s", env))
+			} else {
+				rd.SetId(fmt.Sprintf("file:%s", file))
```
zedkipp, line 40: I suppose we could call `d.SetId(uuid.NewString())` for simplicity here, but this is more descriptive.

provider/secret.go
```diff
@@ -0,0 +1,132 @@
...
+			// Only enforce missing secrets when we are certain this is a
+			// workspace start build. We check both conditions:
+			//  1. CODER_WORKSPACE_BUILD_ID is set (real build, not local
+			//     terraform plan)
+			//  2. CODER_WORKSPACE_TRANSITION is "start"
+			// In all other cases (stop, delete, local dev, ambiguous state)
+			// we return an empty value so the operation can proceed. This
+			// prevents a missing or deleted secret from making a workspace
+			// unstoppable or undeletable.
+			if value == "" {
```
johnstcn, line 61: suggestion: invert the condition here - the "ok" case is pretty trivial: set the value and return nil. You could apply similar logic on line 64 as well: if build ID unset and not a start, return empty value. This should make the overall flow easier to read.

johnstcn (approved):
Looks good to me! Just one note regarding the control flow in the data source.

provider/secret.go
```diff
@@ -0,0 +1,132 @@
...
+				}
+			}
+			_ = rd.Set("value", value)
+
+			return nil
+		},
+		Schema: map[string]*schema.Schema{
+			"env": {
+				Type:        schema.TypeString,
+				Description: "The environment variable name that this secret must inject (e.g. \"MY_TOKEN\"). Must be POSIX-compliant: start with a letter or underscore, followed by letters, digits, or underscores. Exactly one of `env` or `file` must be set.",
```
dylanhuff-at-coder: Are we enforcing POSIX compliance anywhere here? Or enforcing file path rules? I think env and file should both have the same validations done here as the env and file do during user secret creation or update. Without those validations here as well, I believe a template could be made that requires secrets which can't be satisfied since the template could require an env such as `PATH` or a file such as `bad/path` both of which would be rejected when created using the CLI or API

provider/secret.go
zedkipp: I originally tried to make the error path indented, but the readability is much better with the early returns. Done!

dylanhuff-at-coder: I think we should document somewhere that empty secrets will be treated as missing if a template requires it since users can set env and file to "". The behavior makes sense overall, but probably worth explicitly noting it

provider/secret.go
```diff
@@ -0,0 +1,139 @@
+package provider
```
dylanhuff-at-coder, line 1: I think docs are missing from this PR, when I cloned this and ran `make gen`, `secrets.md` was created

provider/secret.go
zedkipp: I was originally hesitant to add any path/env var validation here, mainly due to how validation could skew between coderd and the terraform provider. Even if the requirement can't be satisfied, I think any basic testing by the template author would catch the issue. It just wouldn't be caught earlier at template build time.

That said, I think we can be confident in validating things that are very unlikely to change (e.g. env var names are POSIX compliant). I'm less confident in validating the file paths beyond starting with `~/` or `/` currently.

I'll add basic validation here.

provider/secret.go
zedkipp: Nice catch!! I wonder why this wasn't caught by the CI linting, which seems to explicitly be checking generated docs. I'll check in the updated docs. I also created https://linear.app/codercom/issue/PLAT-142 to fix the CI check

provider/secret.go
zedkipp: This is a good point. I think this behavior is a good thing because it will prevent a user who has a secret with empty `env` and `file` from thinking they have what's needed to create their workspace. I'll add some explicit docs for this.

dylanhuff-at-coder (approved):

PR #507, opened 2026-05-04
chore: warn when coder_agent dir breaks Desktop file sync

zenithwolf1000:
Adds a warning to the `dir` attribute on the `coder_agent` resource for Coder Desktop file sync compatibility.

Two mechanisms:

1. **Doc callout**: A `~> Warning` block in the `dir` schema description, rendered on the Terraform Registry.
2. **Runtime diagnostic**: A `diag.Warn` in `CreateContext` and `ReadWithoutTimeout` that fires during `terraform plan`/`apply` when `dir` is set to a non-`$HOME` value. This warning flows through Terraform output into Coder build logs.

Refs DEVEX-227

> Generated by Coder Agents

### DIFF
--- a/docs/resources/agent.md
+++ b/docs/resources/agent.md
@@ -75,9 +75,9 @@ resource "kubernetes_pod" "dev" {
 - `api_key_scope` (String) Controls what API routes the agent token can access. Options: `all` (full access) or `no_user_data` (blocks `/external-auth`, `/gitsshkey`, and `/gitauth` routes)
 - `auth` (String) The authentication type the agent will use. Must be one of: `"token"`, `"google-instance-identity"`, `"aws-instance-identity"`, `"azure-instance-identity"`.
 - `connection_timeout` (Number) Time in seconds until the agent is marked as timed out when a connection with the server cannot be established. A value of zero never marks the agent as timed out.
-- `dir` (String) The starting directory when a user creates a shell session. Defaults to `"$HOME"`.
+- `dir` (String, Deprecated) The starting directory when a user creates a shell session. Defaults to `"$HOME"`.
 
-~> **Warning:** Setting `dir` to a value other than `$HOME` will break [Coder Desktop file sync](https://coder.com/docs/user-guides/desktop/desktop-connect-sync).
+~> **Warning:** This attribute is deprecated and will be removed in a future release. Setting `dir` to a value other than `$HOME` will break [Coder Desktop file sync](https://coder.com/docs/user-guides/desktop/desktop-connect-sync).
 - `display_apps` (Block Set, Max: 1) The list of built-in apps to display in the agent bar. (see [below for nested schema](#nestedblock--display_apps))
 - `env` (Map of String) A mapping of environment variables to set inside the workspace.
 - `metadata` (Block List) Each `metadata` block defines a single item consisting of a key/value pair. This feature is in alpha and may break in future releases. (see [below for nested schema](#nestedblock--metadata))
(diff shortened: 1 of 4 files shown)


johnstcn (approved):

johnstcn (commented):
Suggestion: you could also add a `diag.Warn` in `agentResource()` so that this will show up in the Terraform log output. May be too noisy though.

code-asher:
>  so that this will show up in the Terraform log output. May be too noisy though.

Ooo that sounds good, and I think that would obviate the need for https://github.com/coder/coder/pull/24951?  Feels to me like having the warning in the provider makes more sense.

matifali:
Should we also mark this as a deprecated attribute?

zenithwolf1000:
I wasn't sure if we wanna fully deprecate the attribute as part of this ticket, was expecting we would have a separate push to properly deprecate and clean up which will take a couple of releases. @matifali I can adjust the warning to say it's deprecated as well if you want

matifali:
Deprecate means we start emitting a warning in terraform build logs, but it will keep working until next major release when we remove it. <br>I think there is a field you can set and it should take care of that.

zenithwolf1000:
Okay, will update to add a deprecation warning too

johnstcn (approved):

mafredri:
I'm happy about the deprecation of this attribute, but it's potentially a pretty big breaking change for DX. Currently we use `dir` in our dogfood template and I've had to change my mental model for `ssh`, `scp` and `rsync` when working in coder. While this fact alone makes it clear that the behavior of `dir` is actively harmful and non-unixy, deprecating it somewhat silently seems like the wrong approach. We should convey to customers how they can migrate and also think about what else may break.

Re: breakage, for instance in Coder Agents `dir` is currently load-bearing for chats being able to automatically include project-level `AGENTS.md`, `.mcp.json`, and skills.

//cc @matifali @johnstcn

matifali:
This is the first phase where we started emitting warnings to discourage it's use. It should still work the same as before. It's a good chance to talk about possible alternatives here.

`dir` was overloaded which we are here trying to solve. I didn't consider that agents also use it.

matifali:
Also this repo need more verbose agent instructions to check any use of attributes in the coder/coder repo too so LLMs can have more context to suggest better completions.

mafredri:
@matifali I think it isn't a great user/customer experience for us to deprecate it without making it clear what our future plans are. Will we support the use-case(s) through some other name? Will we support part of the use-cases, or none at all? Will we split it out into multiple parameters? Customers updating the provider will wonder what they should do now that it's deprecated, and they can't find any answers. To make matters worse the reasoning can't even be discovered in the changelog as this is riding behind the "desktop file sync warning".

matifali:
These are all great points. <br>The two use cases we had in mind when this was being done was

1. Used in Coder Desktop for Filesync. (Which was broken adn we are trying to avoid that)
2. Set `folder` for the builtin`vscode` `display_apps` of `coder_agent`. 

The recommendation for 2, is to use our modules from registry.coder.com which allows explicitly setting `folder`<br>For 1. It is to just not set `dir`at all or set it to `$HOME`.<br><br>I wasn;t aware of any other use cases of this thats whay I said above that this semmes overlaoded attribute.<br><br>@zenithwolf1000 suggsted above to add a follow up instrcutions and deprecation of case 2 [here](<https://linear.app/codercom/review/chore-warn-when-coder-agent-dir-breaks-desktop-file-sync-6a572b3ef90d#comment-8be9ec21>).

PR #538, opened 2026-08-24
docs(provider/env): clarify merge_strategy does not include host environment

bpmct:
The `merge_strategy` schema description for `coder_env` only explained how values merge across `coder_env` resources. It never mentioned that this excludes any value already present in the workspace's environment, such as a base image's `PATH`. A user hit this in coder/coder#27501: appending to `PATH` across three `coder_env` resources replaced the host's `PATH` entirely.

Update the schema description and the `coder_env` example to show referencing `$PATH` in one resource so the host's original value is kept. Ran `make gen` to regenerate `docs/resources/env.md`.

Refs coder/coder#27501

---
*Opened by Coder Agents on behalf of @bpmct.*


provider/env.go
```diff
@@ -47,7 +47,7 @@ func envResource() *schema.Resource {
...
+				Description: "Controls how this environment variable is merged when multiple coder_env resources define the same name. `replace` (default): last value wins. `append`: appends to existing value with a colon `:` separator. `prepend`: prepends to existing value with a colon `:` separator. `error`: fail the build if another coder_env defines the same name. When multiple resources append or prepend to the same name, they are applied in alphabetical order by Terraform resource address. This only merges values across coder_env resources; it does not consider any value already present in the workspace's environment, such as a base image's PATH. To extend such a value, reference it directly, e.g. `value = \"$PATH:/usr/local/bin\"`, so the agent expands it against the real environment at startup.",
```
nickvigilante:
```suggestion
				Description: "Controls how this environment variable is merged when multiple coder_env resources define the same name.\nWhen multiple resources append or prepend to the same name, they are applied in alphabetical order by Terraform resource address.\nThis only merges values across `coder_env` resources; it does not consider any value already present in the workspace's environment, such as a base image's `PATH`.\nTo extend such a value, reference it directly (e.g., `value = \"$PATH:/usr/local/bin\"`), so the agent expands it against the real environment at startup. Valid values are as follows:\n  - `replace` (default): last value wins.\n  - `append`: appends to existing value with a colon `:` separator.\n  - `prepend`: prepends to existing value with a colon `:` separator.\n  - `error`: fail the build if another coder_env defines the same name.",
```

nickvigilante (approved):
I pushed a commit with a few tweaks for clarity and a deliberate sublist for each of the possible values, but LGTM! :shipit: 🚀

PR #540, opened 2026-09-01
fix(provider): validate coder_app URL scheme for external apps

### LINKED ISSUE #483: Validate `coder_app` URL field to require a parseable scheme
Part of [https://github.com/coder/coder/issues/22349](<https://github.com/coder/coder/issues/22349>)

## Problem

The `url` field on the `coder_app` resource has no validation (unlike `icon`, which uses `helpers.ValidateURL`). This allows template authors to specify invalid URLs like bare strings (e.g. `url = "my-repo"`), which are accepted by Go's permissive `url.Parse()` but crash the Coder frontend when JavaScript's stricter `new URL()` constructor encounters them.

## Context

* Go's `url.Parse("my-repo")` succeeds — it treats it as a relative URL
* JavaScript's `new URL("my-repo")` throws `TypeError: Invalid URL` — it requires an absolute URL with a scheme
* The existing `helpers.ValidateURL` also uses `url.Parse` and would **not** catch this class of bug

## Proposed Fix

Add validation to the `url` field that ensures the value would be parseable in a browser/JavaScript context:

1. **When** `external = true`**:** The URL must contain a scheme (e.g. `http://`, `https://`, `vscode://`, `jetbrains-gateway://`, etc.). A bare string or relative path should be rejected.
2. **When** `external = false`**:** The URL is proxied internally and should follow the `http://localhost:PORT[/SUBPATH]` or `http://[::1]:PORT[/SUBPATH]` pattern. Relative URLs without a scheme should still be accepted here since they're resolved server-side.

The validation should go beyond `url.Parse` — at minimum, check that `url.Parse(val).Scheme` is non-empty for external apps. Consider also validating against the set of known/allowed protocols if appropriate.

This is a **breaking change** for templates that currently have invalid URLs, so it may warrant a deprecation warning period or a `terraform plan` diagnostic rather than a hard error.

---

Created on behalf of @angrycub

mmustafasenoglu:
Closes #483

## Summary

Adds URL scheme validation for the `coder_app` resource's `url` field when `external = true`.

## Motivation

Go's `url.Parse()` is permissive and accepts bare strings like `"my-repo"` as relative URLs. However, JavaScript's `new URL()` in the Coder frontend requires a scheme and throws `TypeError: Invalid URL` when encountering these values, crashing the UI.

## Changes

- `provider/app.go`: Added `CustomizeDiff` that checks `url` has a scheme when `external = true`. Internal URLs (external=false) still accept relative paths since they are resolved server-side.
- `provider/helpers/validation.go`: Added `ValidateExternalURL` helper for standalone URL scheme validation.
- `provider/helpers/validation_test.go`: Added unit tests for `ValidateExternalURL`.
- `provider/app_test.go`: Added `ExternalURLValidation` test cases covering external/internal URLs with various schemes.

## Testing

All existing tests pass. New test cases cover:
- External with scheme (https://, vscode://, jetbrains-gateway://) → pass
- External bare string ("my-repo") → error
- External relative path ("/some/path") → error
- Internal bare string / relative path → pass (resolved server-side)

### DIFF
--- a/provider/helpers/validation.go
+++ b/provider/helpers/validation.go
@@ -21,6 +21,36 @@ func ValidateURL(value any, label string) ([]string, []error) {
 	return nil, nil
 }
 
+// ValidateExternalURL validates that a URL intended for external use contains
+// a scheme (e.g. http://, https://, vscode://, jetbrains-gateway://).
+// Go's url.Parse is permissive and accepts bare strings like "my-repo" as
+// relative URLs, but JavaScript's new URL() requires a scheme and will crash
+// if one is not present.
+func ValidateExternalURL(value any, label string) ([]string, []error) {
+	val, ok := value.(string)
+	if !ok {
+		return nil, []error{fmt.Errorf("expected %q to be a string", label)}
+	}
+
+	if val == "" {
+		return nil, nil
+	}
+
+	parsed, err := url.Parse(val)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	if parsed.Scheme == "" {
+		return nil, []error{fmt.Errorf(
+			"%q must have a URL scheme (e.g. https://, vscode://, jetbrains-gateway://), got %q",
+			label, val,
+		)}
+	}
+
+	return nil, nil
+}
+
 // WarnDirNotHome returns a warning if dir is set to a value other
 // than $HOME, because this breaks Coder Desktop file sync. The dir
 // attribute is deprecated and will be removed in a future release.
--- a/provider/app.go
+++ b/provider/app.go
@@ -2,6 +2,8 @@ package provider
 
 import (
 	"context"
+	"fmt"
+	"net/url"
 	"regexp"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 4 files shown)


mmustafasenoglu:
I have read the CLA Document and I hereby sign the CLA

mmustafasenoglu:
recheck

provider/helpers/validation.go
```diff
@@ -21,6 +21,36 @@ func ValidateURL(value any, label string) ([]string, []error) {
 	return nil, nil
 }
 
+// ValidateExternalURL validates that a URL intended for external use contains
+// a scheme (e.g. http://, https://, vscode://, jetbrains-gateway://).
+// Go's url.Parse is permissive and accepts bare strings like "my-repo" as
+// relative URLs, but JavaScript's new URL() requires a scheme and will crash
+// if one is not present.
+func ValidateExternalURL(value any, label string) ([]string, []error) {
```
untra, line 29: label is only used as a detail added to the error message, all the tests only pass "url" as a label . its unneeded as input to this function .

all first param returns are nil, I think these were meant to represent warnings but there are none. neither return parameter needs to be an array . mostly good tests tho. 

could also check that `parsed.Host` is not empty .

untra (changes requested):
Thank you for the contribution. I am also actively working on this issue right now. there are improvements to validation and implementation, but good checking extra beyond `url.Parse(value)` ; external urls should carry a scheme and host.

PR #541, opened 2026-09-01
fix(provider): validate coder_app URL scheme for external apps

mmustafasenoglu:
## Summary

Validate `coder_app.url` when `external = true` so external apps require both a URL scheme and a host.

## Changes

- reject external app URLs missing a scheme or host
- add regression tests for missing-scheme and missing-host inputs

## Validation

- `go test ./provider/...`

### DIFF
--- a/provider/app.go
+++ b/provider/app.go
@@ -2,6 +2,8 @@ package provider
 
 import (
 	"context"
+	"fmt"
+	"net/url"
 	"regexp"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


mmustafasenoglu:
I have read the CLA Document and I hereby sign the CLA

mmustafasenoglu:
Thanks — addressed the edge cases: empty `url` values are rejected when `external = true`, and we validate the parsed hostname (not just `Host`) so values like `https://:8080` are rejected. Added regression tests for empty and missing-host cases.

mmustafasenoglu:
Thanks — addressed the edge cases: empty `url` values are rejected when `external = true`, and we validate the parsed hostname (not just `Host`) so values like `https://:8080` are rejected. Added regression tests for empty and missing-host cases.

PR #542, opened 2026-09-01
fix: external urls validate a scheme and hostname

untra:
* [DEVEX-523](https://linear.app/codercom/issue/DEVEX-523/validate-coder-app-url-field-to-require-a-parseable-scheme)

## Summary
  - coder_app resources with `external = true` now validate the url at plan time via a new `CustomizeDiff` in
    `provider/app.go`, ensuring the URL has both a scheme and a host before apply.
  - Adds `ValidateExternalURL` in `provider/helpers/validation.go`, matching what a browser or OS URL
    handler would actually accept.  Rejects bare host:port strings (which Go's url.Parse misreads as a
    scheme), opaque schemes and malformed urls
    - Increases the timeout of the integration tests for startup in a container 2seconds. this can be controversial; some folks like this tighter.

  ## Why
  External app URLs are opened directly by the browser or a custom protocol handler (e.g. `vscode://`), so
  a URL missing a scheme or missing a host (`https://`) silently fails to open for the user. This catches that at terraform plan time with a clear error instead of a confusing runtime failure.

  Testing
  - New table-driven tests in `provider/helpers/validation_test.go` covering valid cases (`https, vscode://`,
    reverse-DNS custom schemes, `jetbrains-gateway://`) and invalid cases (host:port without scheme,
    opaque schemes, missing scheme, missing host, malformed URLs).
  - New acceptance-style test cases in `provider/app_test.go` (NoScheme, HostPortWithoutScheme, NoHost,
    NotExternalIsUnchecked) verifying the diff-time check only fires when external = true and unknown values (computed at apply) are skipped.
  - `go test ./provider/... passes.`
  - Increases the integration test timeout to 12 seconds from 10; was causing ci flake.
  
  tested locally with: 
  `~/.terraformrc` :
 ```hcl
 terraform {
  required_providers {
    coder = {
      source = "coder/coder"
    }
  }
}

provider "coder" {}

resource "coder_agent" "dev" {
  os   = "linux"
  arch = "amd64"
}

resource "coder_app" "bad_external" {
  agent_id     = coder_agent.dev.id
  slug         = "bad-external"
  display_name = "Bad External App"
  url          = "localhost:8080"
  external     = true
}
 ```
 
 `main.tf` (intentionally broken missing scheme) :
```hcl
  terraform {
  required_providers {
    coder = {
      source = "coder/coder"
    }
  }
}

provider "coder" {}

resource "coder_agent" "dev" {
  os   = "linux"
  arch = "amd64"
}

resource "coder_app" "bad_external" {
  agent_id     = coder_agent.dev.id
  slug         = "bad-external"
  display_name = "Bad External App"
  url          = "localhost:8080"
  external     = true
}
  ```

### DIFF
--- a/provider/app_test.go
+++ b/provider/app_test.go
@@ -109,6 +109,76 @@ func TestApp(t *testing.T) {
 			}
 			`,
 			external: true,
+		}, {
+			name: "NoScheme",
+			config: `
+			provider "coder" {}
+			resource "coder_agent" "dev" {
+				os = "linux"
+				arch = "amd64"
+			}
+			resource "coder_app" "test" {
+				agent_id = coder_agent.dev.id
+				slug = "test"
+				display_name = "Testing"
+				url = "my-repo"
+				external = true
+			}
+			`,
+			external:    true,
+			expectError: regexp.MustCompile(`must include a scheme`),
+		}, {
+			name: "HostPortWithoutScheme",
+			config: `
+			provider "coder" {}
+			resource "coder_agent" "dev" {
+				os = "linux"
+				arch = "amd64"
+			}
+			resource "coder_app" "test" {
+				agent_id = coder_agent.dev.id
+				slug = "test"
+				display_name = "Testing"
+				url = "localhost:8080"
+				external = true
+			}
+			`,
+			external:    true,
+			expectError: regexp.MustCompile(`"localhost" URLs must include a host`),
+		}, {
+			name: "NoHost",
+			config: `
+			provider "coder" {}
+			resource "coder_agent" "dev" {
+				os = "linux"
+				arch = "amd64"
+			}
+			resource "coder_app" "test" {
+				agent_id = coder_agent.dev.id
+				slug = "test"
+				display_name = "Testing"
+				url = "https://"
+				external = true
+			}
+			`,
+			external:    true,
+			expectError: regexp.MustCompile(`"https" URLs must include a host`),
+		}, {
+			name: "NotExternalIsUnchecked",
+			config: `
+			provider "coder" {}
+			resource "coder_agent" "dev" {
+				os = "linux"
+				arch = "amd64"
+			}
+			resource "coder_app" "test" {
+				agent_id = coder_agent.dev.id
+				slug = "test"
+				display_name = "Testing"
+				url = "my-repo"
+			}
+			`,
+			external: false,
 		}}
 		for _, tc := range cases {
 			tc := tc
--- a/provider/helpers/validation.go
+++ b/provider/helpers/validation.go
@@ -34,3 +34,22 @@ func WarnDirNotHome(val interface{}, _ string) ([]string, []error) {
 		`Setting "dir" to a value other than $HOME will break Coder Desktop file sync.`,
 	}, nil
 }
+
+// ValidateExternalURL validates that value is a URL the browser's URL can parse.
+// An external app URL must carry a scheme and hostname
+func ValidateExternalURL(value string) error {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: provider/app_test.go [hunk 1]
```diff
@@ -109,6 +109,76 @@ func TestApp(t *testing.T) {
...
+			resource "coder_app" "test" {
+				agent_id = coder_agent.dev.id
+				slug = "test"
+				display_name = "Testing"
+				url = "localhost:8080"
+				external = true
+			}
+			`,
+			external:    true,
+			expectError: regexp.MustCompile(`"localhost" URLs must include a host`),
```
matifali, line 147: shouldn't this be `"localhost" URLs must include a scheme`

provider/helpers/validation_test.go
```diff
@@ -203,3 +203,117 @@ func TestWarnDirNotHome(t *testing.T) {
...
+			name:  "https URL with path and query",
+			value: "https://example.com/path?q=test",
+		},
+		{
+			name:  "custom vscode scheme",
+			value: "vscode://coder.coder-remote/open?token=$SESSION_TOKEN",
+		},
+		{
+			name:  "reverse DNS scheme with a path",
+			value: "com.example.app://callback",
```
matifali, line 228: is this a valid one or invalid?

provider/helpers/validation_test.go
```diff
@@ -203,3 +203,117 @@ func TestWarnDirNotHome(t *testing.T) {
...
+		},
+		{
+			name:  "jetbrains gateway",
+			value: "jetbrains-gateway://connect#provider=ssh&host=192.168.1.1&port=22&user=ubuntu&projectPath=/home/ubuntu/projects/my-app",
+		},
```
matifali, line 233: Nit: JetBrains Gateway may soon go EOL so this will be corpus in code, so maybe choose something more recent. I suggest replacing it with `antigravity-ide://`

provider/helpers/validation_test.go
```diff
@@ -203,3 +203,117 @@ func TestWarnDirNotHome(t *testing.T) {
...
+		{
+			name:          "protocol relative",
+			value:         "//example.com",
+			errorContains: "must include a scheme",
+		},
+		{
+			name:          "empty string",
+			value:         "",
+			errorContains: "must include a scheme",
```
matifali, line 289: must include a scheme and a host

matifali (approved):
A few nits, but it looks good to me.

provider/helpers/validation.go
```diff
@@ -34,3 +34,22 @@ func WarnDirNotHome(val interface{}, _ string) ([]string, []error) {
 		`Setting "dir" to a value other than $HOME will break Coder Desktop file sync.`,
 	}, nil
 }
+
+// ValidateExternalURL validates that value is a URL the browser's URL can parse.
+// An external app URL must carry a scheme and hostname
+func ValidateExternalURL(value string) error {
```
aslilac, line 40: I think this function probably needs to live in coder/coder somewhere so that...

- we can also validate it on the backend somewhere around here: https://github.com/coder/coder/blob/78b5fdf0ab378f315521c05c6af038e859c3abf5/provisioner/terraform/resources.go#L501
- the validations stay in sync across changes

provider/helpers/validation_test.go
untra: This is technically valid. I was aiming for flexibility on schemes , considering `vscode://`, `zed://`, `jetbrains-gateway://` and other connections down the line; I'm thinking we don't want to maintain a list of allowed schemes here.

